PR #115, opened 2024-07-12
Gracefully handles encoding issues in requirements.txt

rafidka:
*Issue #, if available:* N/A

*Description of changes:* 

The `_requirements_has_constraints` currently fails if the user's `requirements.txt` file is using custom encoding. I updated it such that it uses the `auto_decode` method from `pip` package to make sure we support whatever pip supports.

---

### DIFF
--- a/images/airflow/2.9.2/python/mwaa/entrypoint.py
+++ b/images/airflow/2.9.2/python/mwaa/entrypoint.py
@@ -25,8 +25,7 @@ logging.config.dictConfig(LOGGING_CONFIG)
 
 # Python imports
 from datetime import timedelta
-from functools import cache, partial
-from types import FrameType
+from functools import cache
 from typing import Callable, Dict, List, Optional
 import asyncio
 import json
@@ -66,7 +65,6 @@ from mwaa.subprocess.conditions import (
 from mwaa.subprocess.subprocess import Subprocess, run_subprocesses
 from mwaa.utils.cmd import run_command
 from mwaa.utils.dblock import with_db_lock
-from mwaa.utils.statsd import get_statsd
 
 
 # Usually, we pass the `__name__` variable instead as that defaults to the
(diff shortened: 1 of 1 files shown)


images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -190,15 +190,31 @@ def create_queue() -> None:
...
+        try:
+            with open(requirements_file, "r", encoding=encoding) as file:
+                for line in file:
+                    # Notice that this regex check will also match lines with commented
+                    # out constraints flag. This is intentional as a mechanism for users
+                    # who want to avoid enforcing the default Airflow constraints, yet
+                    # don't want to provide a constraints file.
+                    if re.search(r"-c |--constraint ", line):
+                        return True
+            return False
```
dujjwalr-aws, line 204: If first encoding is not successful, will it not end the loop per-maturely ?

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -190,15 +190,31 @@ def create_queue() -> None:
...
+                f"Failed to decode {requirements_file} using {encoding} encoding."
+            )
 
+    # latin1 is not expected to fail, since it just reads the file as bytes (regardless
+    # of the content), so this block is not expected to be reach, but adding it to be
+    # on the safe side.
+    logging.warning(
+        f"Failed to decode {requirements_file} using {encodings_to_try} encodings. "
+        "Cannot determine whether the requirements file has constraints or not; "
+        "assuming it doesn't."
```
dujjwalr-aws, line 216: if it failed to encode there is no need to add "assuming it doesn't". 
I believe following error message is enough. 
```
Failed to decode {requirements_file} using {encodings_to_try} encodings. "
```

images/airflow/2.9.2/python/mwaa/entrypoint.py
rafidka: No. If the reading fails due to encoding, it will cause an exception, and the next encoding will be tried. If the reading was successful, and we couldn't find the `-c/--constraint` flag, then we return False.

images/airflow/2.9.2/python/mwaa/entrypoint.py
rafidka: Why do you want to remove that part? We were unable to tell if the customer has constraints specified or not, so we ended up making an ** assumption ** that the customer requirements file doesn't have constraints, and this could be incorrect, hence we need to state that clearly in the log to avoid confusion.

dujjwalr-aws: I see what you mean but I feel like "assuming" file context wording is wrong when file decoding failed. we do not know the content of the files so assuming that it doesn't contain constraints is not ideal. 
maybe change of wording will be better.

Mercury2699 (approved):

images/airflow/2.9.2/python/mwaa/entrypoint.py
shubham22: We should just rephrase it "Cannot determine whether the requirements file has constraints or not;" "proceeding without constraining requirements" 
"assuming" is a strange way to say it

shubham22 (changes requested):
Minor update in warning log

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -189,16 +189,36 @@ def create_queue() -> None:
...
+                f"Failed to decode {requirements_file} using {encoding} encoding."
+            )
 
+    # latin1 is not expected to fail, since it just reads the file as bytes (regardless
+    # of the content), so this block is not expected to be reach, but adding it to be
+    # on the safe side.
+    subprocess_logger.warning(
+        f"Failed to decode {requirements_file} using {encodings_to_try} encodings. "
+        "Cannot determine whether the requirements file has constraints or not; "
+        "enforcing Airflow constraints."
```
shubham22, line 219: nit: should we say "enforcing local Airflow constraints"? I think that's what we say in the case no constraint is defined

shubham22 (approved):

images/airflow/2.9.2/python/mwaa/entrypoint.py
rafidka: I think yes, cause this is the action we are taking here when we return False. Unless you are suggesting we don't return True, and hence just enforce the essential constraints?

shubham22: Naah, default action could be what you already have. just add "local" in there : )

jessefrn1 (approved):

dujjwalr-aws (approved):

shubham22 (commented):
LGTM

jessefrn1 (approved):

Mercury2699 (approved):

shubham22 (approved):

PR #119, opened 2024-07-18
Add hybrid scheduler/worker container

kuyperse:
Adds the command 'hybrid' to the entrypoint script to allow creation of a container that runs both celery worker and scheduler sub-processes.

*Issue #, if available:* N/A

*Description of changes:*

Introduces a new supported command 'hybrid' to run celery worker and scheduler sub-processes in the same docker container. Health-check conditions for the 'hybrid' container are limited to the 'worker ' sub-process because the Sidecar does not support multiple health-check simultaneous connections at this time. This 'hybrid' container uses the `CeleryExecutor`, but we plan to add a separate container command to support a `LocalExecutor` setup in the future.

---

### DIFF
--- a/images/airflow/2.9.2/docker-compose-hybrid.yaml
+++ b/images/airflow/2.9.2/docker-compose-hybrid.yaml
@@ -1,0 +1,128 @@
+version: "3.8"
+
+x-airflow-common: &airflow-common
+  image: amazon-mwaa-docker-images/airflow:2.9.2-dev
+  restart: always
+  environment:
+    # AWS credentials
+    AWS_ACCESS_KEY_ID: ${AWS_ACCESS_KEY_ID}
+    AWS_SECRET_ACCESS_KEY: ${AWS_SECRET_ACCESS_KEY}
+    AWS_SESSION_TOKEN: ${AWS_SESSION_TOKEN}
+    AWS_REGION: "us-west-2"
+    AWS_DEFAULT_REGION: "us-west-2"
+
+    # > MWAA Configurations
+    # >> Core Configuration
+    # Use this environment variable if you have custom requirements you want to install when the
+    # container is up.
+    MWAA__CORE__REQUIREMENTS_PATH: ${MWAA__CORE__REQUIREMENTS_PATH}
+    MWAA__CORE__AUTH_TYPE: "testing"
+    # Additional Airflow configuration can be passed here in JSON form.
+    MWAA__CORE__CUSTOM_AIRFLOW_CONFIGS: "{}"
+    MWAA__CORE__FERNET_KEY: '{"FernetKey": "fake-key-nNge+lks3RBeGVrnZ1Dq5GjKerbZKmb7dXNnsNsGy3E="}'
+    MWAA__WEBSERVER__SECRET: '{"secret_key": "fake-key-aYDdF6d+Fjznai5yBW63CUAi0IipJqDHlNSWIun6y8o="}'
+    # Use this enviornment variable to enable encryption with KMS.
+    MWAA__CORE__KMS_KEY_ARN: ${MWAA__CORE__KMS_KEY_ARN}
+    # >> Database configuration
+    MWAA__DB__CREDENTIALS: '{"username": "airflow", "password": "airflow"}'
+    MWAA__DB__POSTGRES_DB: "airflow"
+    MWAA__DB__POSTGRES_HOST: "postgres"
+    MWAA__DB__POSTGRES_PORT: "5432"
+    MWAA__DB__POSTGRES_SSLMODE: "prefer"
+    # >> SQS configuration
+    MWAA__SQS__CREATE_QUEUE: True
+    MWAA__SQS__CUSTOM_ENDPOINT: http://sqs:9324
+    MWAA__SQS__QUEUE_URL: http://sqs:9324/000000000000/celery-queue
+    MWAA__SQS__USE_SSL: False
+    # >> Logging configuration.
+    # Use these environment variables if you want to publish logs to CloudWatch.
+    MWAA__LOGGING__AIRFLOW_DAGPROCESSOR_LOGS_ENABLED: ${MWAA__LOGGING__AIRFLOW_DAGPROCESSOR_LOGS_ENABLED}
+    MWAA__LOGGING__AIRFLOW_DAGPROCESSOR_LOG_GROUP_ARN: ${MWAA__LOGGING__AIRFLOW_DAGPROCESSOR_LOG_GROUP_ARN}
+    MWAA__LOGGING__AIRFLOW_DAGPROCESSOR_LOG_LEVEL: ${MWAA__LOGGING__AIRFLOW_DAGPROCESSOR_LOG_LEVEL}
+    MWAA__LOGGING__AIRFLOW_SCHEDULER_LOGS_ENABLED: ${MWAA__LOGGING__AIRFLOW_SCHEDULER_LOGS_ENABLED}
+    MWAA__LOGGING__AIRFLOW_SCHEDULER_LOG_GROUP_ARN: ${MWAA__LOGGING__AIRFLOW_SCHEDULER_LOG_GROUP_ARN}
+    MWAA__LOGGING__AIRFLOW_SCHEDULER_LOG_LEVEL: ${MWAA__LOGGING__AIRFLOW_SCHEDULER_LOG_LEVEL}
+    MWAA__LOGGING__AIRFLOW_TRIGGERER_LOGS_ENABLED: ${MWAA__LOGGING__AIRFLOW_TRIGGERER_LOGS_ENABLED}
+    MWAA__LOGGING__AIRFLOW_TRIGGERER_LOG_GROUP_ARN: ${MWAA__LOGGING__AIRFLOW_TRIGGERER_LOG_GROUP_ARN}
+    MWAA__LOGGING__AIRFLOW_TRIGGERER_LOG_LEVEL: ${MWAA__LOGGING__AIRFLOW_TRIGGERER_LOG_LEVEL}
+    MWAA__LOGGING__AIRFLOW_TASK_LOGS_ENABLED: ${MWAA__LOGGING__AIRFLOW_TASK_LOGS_ENABLED}
+    MWAA__LOGGING__AIRFLOW_TASK_LOG_GROUP_ARN: ${MWAA__LOGGING__AIRFLOW_TASK_LOG_GROUP_ARN}
+    MWAA__LOGGING__AIRFLOW_TASK_LOG_LEVEL: ${MWAA__LOGGING__AIRFLOW_TASK_LOG_LEVEL}
+    MWAA__LOGGING__AIRFLOW_WEBSERVER_LOGS_ENABLED: ${MWAA__LOGGING__AIRFLOW_WEBSERVER_LOGS_ENABLED}
+    MWAA__LOGGING__AIRFLOW_WEBSERVER_LOG_GROUP_ARN: ${MWAA__LOGGING__AIRFLOW_WEBSERVER_LOG_GROUP_ARN}
+    MWAA__LOGGING__AIRFLOW_WEBSERVER_LOG_LEVEL: ${MWAA__LOGGING__AIRFLOW_WEBSERVER_LOG_LEVEL}
+    MWAA__LOGGING__AIRFLOW_WORKER_LOGS_ENABLED: ${MWAA__LOGGING__AIRFLOW_WORKER_LOGS_ENABLED}
+    MWAA__LOGGING__AIRFLOW_WORKER_LOG_GROUP_ARN: ${MWAA__LOGGING__AIRFLOW_WORKER_LOG_GROUP_ARN}
+    MWAA__LOGGING__AIRFLOW_WORKER_LOG_LEVEL: ${MWAA__LOGGING__AIRFLOW_WORKER_LOG_LEVEL}
+
+  volumes:
+    - ./dags:/usr/local/airflow/dags
+    - ./plugins:/usr/local/airflow/plugins
+    - ./requirements:/usr/local/airflow/requirements
+    - ./startup:/usr/local/airflow/startup
+
+  depends_on: &airflow-common-depends-on
+    postgres:
+      condition: service_healthy
+    sqs:
+      condition: service_healthy
+
+services:
+  postgres:
+    image: postgres:13
+    container_name: mwaa-292-db
+    environment:
+      POSTGRES_USER: airflow
+      POSTGRES_PASSWORD: airflow
+      POSTGRES_DB: airflow
+    volumes:
+      - postgres-db-volume:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "airflow"]
+      interval: 10s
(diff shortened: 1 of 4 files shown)


images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -610,6 +615,16 @@ def run_airflow_command(cmd: str, environ: Dict[str, str]):
...
 
+        # Hybrid runs the scheduler and celery worker processes is a single container.
+        case "hybrid":
+            # The Sidecar healthcheck is currently limited to one healthcheck per port
+            # so the hybrid container can only include healthchecks for one subprocess.
+            # Only the worker healcheck conditions are enabled to monitor container health, so
+            # we pass an empty list of conditions to the scheduler process.
+            scheduler_subprocesses = _get_airflow_scheduler_subprocesses(environ, [])
+            worker_subprocesses = _get_airflow_worker_subprocesses(environ)
+            run_subprocesses(worker_subprocesses + scheduler_subprocesses)
```
Kytha, line 626: Since the worker subprocess has the health conditions attached, it needs to be made an essential process. Otherwise, if the worker exits due to failed healthcheck and scheduler is still running then the task will not be restarted but worker is lost.

images/airflow/2.9.2/docker-compose-hybrid.yaml
```diff
@@ -0,0 +1,128 @@
+version: "3.8"
```
dhegberg, line 1: Since so much of this is duplicated from the [default compose](https://github.com/aws/amazon-mwaa-docker-images/blob/main/images/airflow/2.9.2/docker-compose.yaml) can we move everything shared to a base compose file and [extend](https://docs.docker.com/compose/multiple-compose-files/extends/) for the split/hybrid worker options

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -496,19 +498,8 @@ def _get_sidecar_health_port():
...
-    if _is_sidecar_health_monitoring_enabled():
-        conditions.append(
-            SidecarHealthCondition(
-                airflow_component="worker",
-                container_start_time=CONTAINER_START_TIME,
-                port=_get_sidecar_health_port(),
-            ),
-        )
-
+def _get_airflow_worker_subprocesses(environ: Dict[str, str]):
```
rafidka, line 501: `create` is a more appropriate verb here than `get`. The latter is typically used when you are retrieving a particular thing, rather than creating a new one every time the method is called.

Also, please add a docstring to this method.

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -548,9 +538,51 @@ def on_sigterm() -> None:
                 on_sigterm=on_sigterm,
             )
         ]
-    )
 
 
+def _get_airflow_scheduler_subprocesses(environ: Dict[str, str], conditions: List):
```
rafidka, line 543: Like above, please use `create` instead of `get`.

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -548,9 +538,51 @@ def on_sigterm() -> None:
...
+            )
+            for airflow_cmd, logger_name, friendly_name in [
+                ("scheduler", SCHEDULER_LOGGER_NAME, "scheduler"),
+                ("dag-processor", DAG_PROCESSOR_LOGGER_NAME, "dag-processor"),
+                ("triggerer", TRIGGERER_LOGGER_NAME, "triggerer"),
+            ]
+        ]
+
+
+def _get_conditions(airflow_cmd: str):
```
rafidka, line 567: `_get_conditions` is an ambiguous name. What conditions are we getting? I would rather use a long, slightly ugly, but informative name like `get_scheduler_worker_common_conditions` than an ambiguous name. Ideally though, an informative and beautiful name would be great.

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -610,6 +615,17 @@ def run_airflow_command(cmd: str, environ: Dict[str, str]):
                 ]
             )
 
+        # Hybrid runs the scheduler and celery worker processes is a single container.
+        case "hybrid":
```
rafidka, line 619: I don't particularly like the name "hybrid". Hybrid of what? In our offline discussions, we pretty much all understood what the term hybrid mean, but from the perspective of this code, it doesn't mean much. For example, if I run the following command:

```
docker run amazon-mwaa-docker-images:2.9.2 worker
```

it is clear what is being run. On other hand, no one will understand what this means:

```
docker run amazon-mwaa-docker-images:2.9.2 hybrid
```

rafidka (changes requested):
Looks good in general, just left a couple cosmetic comments.

But also, it is about time we start adding tests to this repo. The initial code was written during the work on supporting 2.9.2, and thus I had limited time to work on testing. But from now on, we should start adding testing for new code.

images/airflow/2.9.2/python/mwaa/entrypoint.py
rafidka: Regardless of health conditions, any process can die for any reason, so yes the scheduler and worker should both be considered essential processes, to ensure that if one fails, the other also exits so we get the container restarted.

Having said that, the implication of this is that if the scheduler fails for any reason, it will result in forcefully terminating the worker within [90 seconds](https://github.com/aws/amazon-mwaa-docker-images/blob/main/images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py#L31-L33), which can result in task interruption. Thus, at the least we should:

1. Consider increasing the patience period for the worker to give tasks some time to finish. Obviously, we have to be careful how big this value we make it, as the downside is that it means during the patience period, we don't have a scheduler running.
2. Make sure our customers are aware of this.

The ideal solution, though, is to have the different processes restart independent of each other, i.e. doesn't cause a container restart. One idea that I wanted to work on but didn't have the time for is to implement a RetryableSubprocess class (which inherits from the Subprocess class). As the name suggests, this class restart the underlying subprocess if it fails. Then, when you run the workers and schedulers in a hybrid container, you use a RetryableSubprocess instead of Subprocess, meaning that if the scheduler/worker fails, it will retried "individually", up to a certain number of times after which we give up and accept that we will have to terminate running tasks.

Kytha: Good point. I will make both processes essential and increase the patience interval for the worker to 120s. Getting the base functionality implemented here will make end-to-end testing easier, and I can follow up with the implementation of the RetryableSubprocess to improve the experience here.

images/airflow/2.9.2/docker-compose-hybrid.yaml
Kytha: Sure, I'll refactor this

images/airflow/2.9.2/python/mwaa/entrypoint.py
Kytha: I think maybe _get_airflow_conditions would be better. I don't think the name requires the addition of "worker" or "scheduler" since this is essentially a parameter to the function. For the given airflow process command, this function returns the conditions required.

images/airflow/2.9.2/python/mwaa/entrypoint.py
Kytha: I do agree that the name is slightly ambiguous,  do you have a suggestion for what an alternate name should be?

The problem I see is that if we choose an explicit name like `scheduler_worker`, this is clear in isolation, but would typically become confusing in most other contexts. For example `HYBRID_SIGTERM_PATIENCE_INTERVAL` becomes  `SCHEDULER_WORKER_SIGTERM_PATIENCE_INTERVAL` which is arguably much worse than `hybrid`.

And I think `SCHEDULER_<AND/WITH>_WORKER` also has the same issue. So I am struggling to find a better alternative that does not have ambiguity.

images/airflow/2.9.2/docker-compose-hybrid.yaml
Kytha: So turns out docker-compose lost support for extends as of v3. The alternative is to use file overrides like `docker-compose -f "base.yaml" -f "override.yaml"`. However, this breaks the yaml anchor references, unless we redefine them across all files. There is a workaround discussed [here](https://github.com/docker/compose/issues/5621#issuecomment-581797428) by concating the files and piping to docker-compose, but I've not been able to get that to work due to duplicate definitions.  

I think there is a way to achieve what we want, but may require more research and a larger refactor of the compose structure. I would prefer to leave this out of scope for this PR.

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -95,6 +97,8 @@
 # related to sidecar endpoint not reporting health messages.
 CONTAINER_START_TIME = time.time()
 
+HYBRID_WORKER_SIGTERM_PATIENCE_INTERVAL = timedelta(seconds=130)
```
Kalhan22, line 100: Is this any arbitrary value? If not, please add a note on what this means.

Kytha: Yeah a comment would make this more clear. Is there a particular reason we would want this configuration to be set by the container orchestrator? It is not a configuration that is currently dependent on external conditions.

images/airflow/2.9.2/docker-compose-hybrid.yaml
rafidka: Yeah, I think it is fine to leave this for another PR. To be honest, personally, I don't like the limitations of Docker and Docker Compose when it comes to programmability, and this is why for the Dockerfiles, I ended up using a templating engine (Jinja). For Docker Compose, if the need arises for more sophistication, we can think of a solution later, ideally a mechanism that automatically converts Fargate service definitions into Docker Compose, or some DSL that automate the generation of both Fargate services and Docker Compose.

images/airflow/2.9.2/python/mwaa/entrypoint.py
rafidka: `_get_airflow_conditions` is still vague in my opinion. Perhaps `_get_airflow_process_conditions` is better, since it becomes clearer what the conditions are for, and the parameter explains what Airflow process we are talking about.

images/airflow/2.9.2/python/mwaa/entrypoint.py
rafidka: In cases where existing terms are not helping, I sometimes try to come up with a new, nice, and unused term that succinctly describe the new thing. Some examples I can think of:

- `mini_airflow`
- `full_airflow`
- `airflow_pod`

Probably none of these is the best (re 1, it is not necessarily mini, we could have huge compute, re 2 it is not full, it is missing web server, and so on), but I am just trying to give examples here.

In any case, it is not the end of the world if we don't come up with a better name. We can always change later, so I don't want to block you here. So, feel free to resolve the conversation.

images/airflow/2.9.2/python/mwaa/entrypoint.py
rafidka: I would say we could allow overriding the patience interval via an environment variable, but I recommend keeping a default value. In general, we want the user to be able to run our Docker image with the minimal amount of configuration, so unless absolutely necessary, all our configurable parameters should have default values.

In the new images, I have adopted a configuration style similar to that of Airflow. So, if you want to make this overridable by a configuration, you could do something like:

`MWAA__HYBRID_CONTAINER__SIGTERM_PATIENCE_INTERVAL`

rafidka: My comment above was in response to Kalhan's comment.

> Is there a particular reason we would want this configuration to be set by the container orchestrator? It is not a configuration that is currently dependent on external conditions.

Usually, when the value of a certain parameter has an implication, we want to make it configurable so the users can customize it to their needs. Considering that the terminating a worker can potentially result in task interruption, I would say this is a parameter that benefits from being configurable. Having said that, if we don't anticipate that we will allow this to be configurable in MWAA, and we don't anticipate that some of our users might want to use this in a vanilla setup that benefits from configuring this, then I would say we can keep it in code for now.

In any case, I usually prefer to focus on what we need and bias for action, as opposed to trying to prematurely anticipate what we need in the future, and slow ourselves down.

Kalhan22: If this value is only ever going to be used when building a hybrid docker image, then it seems more intuitive to have its intended value passed down by only that docker compose. I'd assume since we used "hybrid" in its name, non hybrid images wouldn't ever need to rely on its value.

If this is going to be indeed needed when building any other docker image, then perhaps it is fine as it is. 

Perhaps the presence of "hybrid" in this variable name is causing confusion on the intent.

images/airflow/2.9.2/python/mwaa/entrypoint.py
Kytha: Yeah, it is something that if we find a better name we can always change later. The problem I see with all our names so far is they still require additional context to understand. For example, both `hybrid` and `airflow_pod` are ambiguous and would require reader having some additional background in what those terms mean in the context of our image.

images/airflow/2.9.2/python/mwaa/entrypoint.py
Kytha: I would argue the opposite. If this value was only supplied by the orchestrator (and the image has no opinion on this config based on the mode), than from MWAA perspective, we would need to supply both the 'hybrid' option *and* the correct patience timeout for the container to behave as it should for hybrid mode. 

To me this would be unintuitive. If I tell the container I want hybrid, the container should configure itself internally to suite that use-case. I shouldn't need to understand or be aware of subprocess patience to deploy a hybrid container correctly. 

However, we can still provide it as a option for the orchestrator to configure for the scenario where the user wants more fine-grain control without modifying the image. However, `MWAA__HYBRID_CONTAINER__SIGTERM_PATIENCE_INTERVAL` value would still need to remain in the image code as fallback because if orchestrator does not supply any value we still want container to behave optimally in hybrid mode.

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -622,9 +634,24 @@ def run_airflow_command(cmd: str, environ: Dict[str, str]):
...
+            # the default patience interval for worker needs to be increased to better
+            # allow for in-flight tasks to complete in case of scheduler failure.
+            try:
+                worker_patience_interval_seconds = os.getenv('MWAA__HYBRID_CONTAINER__SIGTERM_PATIENCE_INTERVAL', None)
+                if worker_patience_interval_seconds is not None:
+                    worker_patience_interval = timedelta(seconds=int(worker_patience_interval_seconds))
+                else:
+                    worker_patience_interval = HYBRID_WORKER_SIGTERM_PATIENCE_INTERVAL_DEFAULT
+            except (ValueError, TypeError):
+                print("BANNA!")
```
dhegberg, line 649: ^

Kytha: My bad, artifact from testing XD

rafidka (approved):

dhegberg (approved):

PR #120, opened 2024-07-19
Add Env. Variables to set CORS config.

NikhilRaverkar:
*Issue #, if available:*
This PR adds an environment variable to allow set CORS config. CORS will be enabled by default unless MWAA__API__ALLOWED_HEADERS, MWAA__API__ALLOWED_METHODS MWAA__API__ALLOWED_ORIGINS  are explicitly set.
*Description of changes:*
Verified the change by setting the AIRFLOW__API__ACCESS_CONTROL_ALLOW_HEADERS, AIRFLOW__API__ACCESS_CONTROL_ALLOW_METHODS and AIRFLOW__API__ACCESS_CONTROL_ALLOW_ORIGINS to * 


images/airflow/2.9.2/python/mwaa/config/airflow.py
```diff
@@ -272,6 +272,14 @@ def _get_essential_airflow_api_config() -> Dict[str, str]:
     if os.environ.get("MWAA__CORE__AUTH_TYPE", "").lower() == "none":
         api_config["AIRFLOW__API__AUTH_BACKENDS"] = "airflow.api.auth.backend.default"
 
+    if os.environ.get("MWAA__API__ALLOWED_HEADERS"):
+        api_config["AIRFLOW__API__ACCESS_CONTROL_ALLOW_HEADERS"] = os.environ.get("MWAA__API__ALLOWED_HEADERS") 
+    if os.environ.get("MWAA__API__ALLOWED_METHODS"):
+        api_config["AIRFLOW__API__ACCESS_CONTROL_ALLOW_METHODS"] = os.environ.get("MWAA__API__ALLOWED_METHODS")
+    if os.environ.get("MWAA__API__ALLOWED_ORIGINS"):
+        api_config["AIRFLOW__API__ACCESS_CONTROL_ALLOW_ORIGINS"] = os.environ.get("MWAA__API__ALLOWED_ORIGINS")
```
rafidka, line 280: We don't need to create a corresponding MWAA configuration for every Airflow configuration. You can simply make use of the [`MWAA__CORE__CUSTOM_AIRFLOW_CONFIGS`](https://github.com/aws/amazon-mwaa-docker-images/blob/main/images/airflow/2.9.2/python/mwaa/config/airflow.py#L96).

rafidka: To elaborate, we have a set of Airflow configuration that are either essential (e.g. Celery config) or we are opinionated about, and so we define them explicitly. Everything else, however, is up to the user to define using the CUSTOM_AIRFLOW_CONFIGS environment variable.

PR #123, opened 2024-07-24
Add dnf update to dockerfile and bootstrap for OS packages

michaelfu1029:
Added a dnf update inside the bootstrap

*Issue #, if available:*
N/A

*Description of changes:*
Added these as recommended for PCI compliance

---

### DIFF
--- a/images/airflow/2.9.2/bootstrap/01-root-firstpass/001-init.sh
+++ b/images/airflow/2.9.2/bootstrap/01-root-firstpass/001-init.sh
@@ -2,3 +2,4 @@
 set -e
 
 dnf update -y
+dnf update --security -y
--- a/images/airflow/2.9.2/bootstrap/01-root-firstpass/100-install-needed-dnf-packages.sh
+++ b/images/airflow/2.9.2/bootstrap/01-root-firstpass/100-install-needed-dnf-packages.sh
@@ -6,4 +6,3 @@ dnf install -y libcurl-devel                                # For pycurl
 dnf install -y postgresql-devel                             # For psycopg2
 dnf install -y procps                                       # For 'ps' command, which is used for monitoring.
 dnf install -y jq                                           # For 'jq' command, which is used in execute-startup.sh
-dnf update --security -y                                    # Update OS packages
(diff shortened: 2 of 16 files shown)


Kalhan22 (dismissed):

images/airflow/2.9.2/bootstrap/01-root-firstpass/100-install-needed-dnf-packages.sh
```diff
@@ -5,4 +5,5 @@ dnf install -y java-17-amazon-corretto                      # For Java lovers.
 dnf install -y libcurl-devel                                # For pycurl
 dnf install -y postgresql-devel                             # For psycopg2
 dnf install -y procps                                       # For 'ps' command, which is used for monitoring.
-dnf install -y jq                                           # For 'jq' command, which is used in execute-startup.sh
\ No newline at end of file
+dnf install -y jq                                           # For 'jq' command, which is used in execute-startup.sh
+dnf update --security -y                                    # Update OS packages
```
rafidka, line 9: I noticed that you are doing the update in two places, here in this file, and in the Dockerfile.base.j2 itself. Some comments here:

1. Is there any reason you need to execute the `update --security` command multiple times? I guess no, but just wanted to check.
2. With such steps, we usually try to avoid directly changing the Dockerfile, and instead just change the bootstrapping steps or add a new one.
3. This file is not the right place for this command (I think I should probably change the name since it seems confusing). You should instead do such fundamental steps in the `001-init.sh` file, as we are doing an update there, so all you need to do is change that line and we should be good.

michaelfu1029: Updated PR to address these comments

Kalhan22 (approved):

images/airflow/2.9.2/bootstrap/01-root-firstpass/001-init.sh
```diff
@@ -2,3 +2,4 @@
 set -e
 
 dnf update -y
+dnf update --security -y
```
rafidka, line 5: You don't actually need to execute the update command twice, just one line of `dnf update --security -y` should be enough. From the help page:

```
  --security            Include security relevant packages, in updates
```

michaelfu1029:
Not needed as discussed

Issue #124, opened 2024-07-26
The serialized_dag table is cleared when every new instance is started


millin:
**Describe the bug**
The `serialized_dag` table is cleared when every new instance (e.g. worker) is started.

This results in the following errors:
- killing a recently started task (zombie)
- DagNotFound exception, when `trigger_dag` is calling
- `DAG <dag_id> seems to be missing from DagBag` error in webserver
- dag-processor error (https://github.com/apache/airflow/issues/40082)

---

The error occurs very often because when starting a new image (e.g. for new worker) each time the `airflow db migrate` [command is called](https://github.com/aws/amazon-mwaa-docker-images/blob/2c0bad6b29f8f856fe4c4938a7b6d99ab626f1ec/images/airflow/2.9.2/python/mwaa/entrypoint.py#L114) from `entrypoint`, which in turn runs `reserialize_dags` and clears the whole `serialized_dag` table.

_Originally posted by @millin in https://github.com/apache/airflow/issues/40082#issuecomment-2253016337_

rafidka:
Thank you, @millin . I will try to take a look today and get back to you with what I find. I am surprised the `serialized_dag` is cleared with every call to `db migrate`. I thought it is basically a no-op in case the DB is already up to date.

millin:
I am also surprised, however the defaults of `reserialize_dags` argument is `True`:
https://github.com/apache/airflow/blob/f0ef69198ec0b7ad0c489cbccf76f6130445fedf/airflow/cli/cli_config.py#L659-L666

I can suggest a quick fix
```diff
--- await run_command("airflow db migrate", env=environ)
+++ await run_command("airflow db migrate --no-reserialize-dags", env=environ)
```
But I'm not sure it won't cause problems when updating the Airflow version

rafidka:
Yeah, I also noticed the `reserialize_dags` argument. In addition to not being super comfortable about using an argument which is [not documented](https://airflow.apache.org/docs/apache-airflow/2.9.2/cli-and-env-variables-ref.html#migrate), I also think the re-serialization is required if we are doing a version upgrade, which is possible with MWAA in case of version upgrade. Ideally, Airflow shouldn't be doing the re-serialization if the DB is already migrated, and we should probably report this as a bug in Airflow. However, for MWAA, we will probably have to change the code to do the check ourselves and avoid calling `db migrate` if the DB is already initialized.

rafidka:
[...]

PR #125, opened 2024-07-29
Fix issue 123: Repeated reserialization of DAGs

rafidka:
*Issue #, if available:* #124

## Description of changes

Issue #124 happens because we are always calling `airflow db migrate` to make sure the database is migrated. The assumption was that the operation is a no-op if the DB is already initialized. However, this is not the case, and a reserialization of DAGs is happening every time this command is called. To avoid this, I am changing the code for executing migration to check if migration is needed.

## Testing

I did multiple tests using the Docker Compose setup. More testing by Amazon MWAA developers before this PR is merged. Suggested testing:

1. Create an internal 2.9.2 environment and make sure it is functioning normally. Also check ECS logs to ensure there are no issues.
2. Create a pre-2.9.2 environment and do an in-place upgrade to 2.9.2 and make sure the process is flawless.
3. Run a 2.9.2 environment with auto scaling enabled (ensuring that workers are being created and terminated continuously) and make sure migrations are not causing any issues.

---

### DIFF
--- a/images/airflow/2.9.2/python/mwaa/database/migrate.py
+++ b/images/airflow/2.9.2/python/mwaa/database/migrate.py
@@ -1,0 +1,69 @@
+"""
+This script is responsible for running Airflow meta database migrations.
+
+IMPORTANT NOTE: This script must be run with all the required environments exported,
+just like when running any Airflow command, as it imports Airflow modules and needs to
+connect to the meta database, thus all configurations need to be set.
+"""
+
+from argparse import Namespace
+import logging.config
+import os
+import sys
+
+from mwaa.utils.dblock import with_db_lock
+
+
+# Usually, we pass the `__name__` variable instead as that defaults to the module path,
+# i.e. `mwaa.entrypoint` in this case. However, since this is a script, `__name__` will
+# have the value of `__main__`, hence we hard-code the module path.
+logger = logging.getLogger("mwaa.database.migrate")
+
+
+def _verify_environ():
+    """
+    This script is supposed to have all the environment variables required for running
+    Airflow, since we will be using Airflow modules directly. This function verifies
+    they are set by ensuring the existence of the `AWS_EXECUTION_ENV`, which we add
+    during the creation of the `environ` dictionary in the entrypoint.py.
+    """
+    if not os.environ.get("AWS_EXECUTION_ENV", "").startswith("Amazon_MWAA_"):
+        logger.error("The necessary environment variables are not set.")
+        sys.exit(1)
+
+
+@with_db_lock(1234)
+def _migrate_db():
+    from airflow.cli.commands import db_command as airflow_db_command
+
+    try:
+        args = Namespace(migration_wait_timeout=1)
+        airflow_db_command.check_migrations(args)
+        logging.info("The database is already migrated.")
+    except TimeoutError:
+        logging.info("The database is not yet migrated. Migrating...")
+        args = Namespace(
+            from_revision=None,
+            from_version=None,
+            reserialize_dags=None,
+            show_sql_only=None,
+            to_revision=None,
+            to_version=None,
+            use_migration_files=None,
+        )
+        airflow_db_command.migratedb(args)
+        logging.info("The database is now migrated.")
+
+
+def _main():
+    _verify_environ()
+    _migrate_db()
+
+
+if __name__ == "__main__":
+    _main()
+else:
+    logger.error(
+        "This module cannot be imported. It should be run directly using: python -m mwaa.database.migrate"
+    )
+    sys.exit(1)
--- a/images/airflow/2.9.2/BillOfMaterials/2.9.2-dev/Python-Packages-BOM.txt
+++ b/images/airflow/2.9.2/BillOfMaterials/2.9.2-dev/Python-Packages-BOM.txt
@@ -5611,10 +5611,10 @@ Copyright 2013-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 /usr/local/airflow/.local/lib/python3.11/site-packages/boto3-1.34.106.dist-info/NOTICE
 
 boto3-stubs
-1.34.143
+1.34.149
 MIT License
 https://github.com/youtype/mypy_boto3_builder
-/usr/local/airflow/.local/lib/python3.11/site-packages/boto3_stubs-1.34.143.dist-info/LICENSE
+/usr/local/airflow/.local/lib/python3.11/site-packages/boto3_stubs-1.34.149.dist-info/LICENSE
 MIT License
 
 Copyright (c) 2024 Vlad Emelianov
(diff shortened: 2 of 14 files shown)


images/airflow/2.9.2/python/mwaa/database/migrate.py
```diff
@@ -0,0 +1,69 @@
...
+        args = Namespace(
+            from_revision=None,
+            from_version=None,
+            reserialize_dags=None,
+            show_sql_only=None,
+            to_revision=None,
+            to_version=None,
+            use_migration_files=None,
+        )
+        airflow_db_command.migratedb(args)
```
dhegberg, line 54: I'm not seeing `migratedb` in the docs, only [`migrate`](https://airflow.apache.org/docs/apache-airflow/2.9.2/cli-and-env-variables-ref.html#migrate)

images/airflow/2.9.2/python/mwaa/database/migrate.py
```diff
@@ -0,0 +1,69 @@
...
+        logger.error("The necessary environment variables are not set.")
+        sys.exit(1)
+
+
+@with_db_lock(1234)
+def _migrate_db():
+    from airflow.cli.commands import db_command as airflow_db_command
+
+    try:
+        args = Namespace(migration_wait_timeout=1)
```
dhegberg, line 40: Is a one second check sufficient or will we have false Timeout errors?  

Is the migrate command is idempotent to handle the case of false Timeout errors?

images/airflow/2.9.2/python/mwaa/database/migrate.py
```diff
@@ -0,0 +1,69 @@
...
+        airflow_db_command.check_migrations(args)
```
dhegberg, line 41: Just to confirm, this command is meant to wait until the migration is done?  So if the database is on a prior version we'd expect a quick timeout, otherwise it has a response we ignore.

images/airflow/2.9.2/python/mwaa/database/migrate.py
dhegberg: If always calling migrate was causing the re-serialization issue, it would seem this is not idempotent.  Is that accurate?

images/airflow/2.9.2/python/mwaa/database/migrate.py
dhegberg: Nevermind, I guess it's this:
https://github.com/apache/airflow/blob/36b92345b91e777f64dd860a740241a89c1232fc/airflow/cli/commands/db_command.py#L98

images/airflow/2.9.2/python/mwaa/database/migrate.py
rafidka: > Is a one second check sufficient or will we have false Timeout errors?

I need a method for checking whether the DB is initialized or not, but Airflow CLI didn't provide that unfortunately, so I had to employ the workaround of using `check_migrations` without a timeout.

> Is the migrate command is idempotent to handle the case of false Timeout errors?

Like above, I am using the check_migrations command to test whether migrations have been done or not, without a timeout.

> If always calling migrate was causing the re-serialization issue, it would seem this is not idempotent. Is that accurate?

Depends on what you mean by "this". Airflow's `db migrate` command is certainly not idempotent. However, the `_migrate_db` command is idempotent, and the reason is that we have it surrounded by `with_db_lock` meaning that no two processes can execute it at the same time, and within each call, we check whether the DB is initialized or not, and only initialize it if it is not.

images/airflow/2.9.2/python/mwaa/database/migrate.py
rafidka: > Just to confirm, this command is meant to wait until the migration is done?

Nope, it is just used as a method to check whether migrations have been applied or not.  See the detailed discussion above. I think I should refactor the code a bit to make it clear we are using check_migrations differently.

images/airflow/2.9.2/python/mwaa/database/migrate.py
rafidka: Exactly. The name in code doesn't match the name in the CLI.

images/airflow/2.9.2/python/mwaa/database/migrate.py
dhegberg: So with 2 (or more) schedulers.  

Previously one would get the lock, run migrate and release the lock.  The other would wait until the lock was released and run migrate resulting in the re-serialization.

Now one would get the lock, check the migration, run migrate and release the lock.  The other would wait until the lock was release and check the migration and exit.

Makes sense.

dhegberg (approved):

images/airflow/2.9.2/python/mwaa/database/migrate.py
ms964107: Ref: https://github.com/apache/airflow/blob/36b92345b91e777f64dd860a740241a89c1232fc/airflow/utils/db.py#L840

ms964107 (approved):

jaklan:
When it's merged, will MWAA 2.9.2-based instances be automatically updated? We are also affected by https://github.com/apache/airflow/issues/40082 and waiting for the fix before we propagate upgrade from 2.8 to 2.9 on PRD

rafidka:
> When it's merged, will MWAA 2.9.2-based instances be automatically updated? We are also affected on DEV & UAT environments by [apache/airflow#40082](https://github.com/apache/airflow/issues/40082) and waiting for the fix before we propagate upgrade from 2.8 to 2.9 on PRD

Unfortunately no, you would need to trigger an environment update to have the Docker containers updated. Even a no-op update would do, though, so it should be pretty easy to do.

jaklan:
@rafidka sure, that's not an issue. Are changes immediately propagated to MWAA after merging PR or is there any build- / deployment-related delay we have to take into consideration before we can trigger an update and test new containers?

rafidka:
> Are changes immediately propagated to MWAA after merging PR or is there any build- / deployment-related delay we have to take into consideration before we can trigger an update and test new containers?

After the PR is merged, we sync it internally and deploy it via internal pipelines where it goes through testing, and then becomes available to different regions around the world gradually. Usually, assuming no test failures or pipeline blockages, this takes a couple of days. I am currently out of office, but @dhegberg should be able to help you with status updates. Daniel, we need to have this PR merged asap, as it is a production issue for our customers.

Mercury2699 (approved):

jaklan:
MR is merged, so asking for the status update then 😄 Can we proceed with the re-deployment or nor yet?

shubham22:
@Mercury2699 - can you confirm if this has been deployed to all the waves/regions?

Mercury2699:
@jaklan Yes, this fix has been deployed to all regions. Customers can trigger environment update to receive the latest image.

PR #133, opened 2024-08-27
Fix duplicated scheduler processes + make all airflow processes essen…

Kytha:
*Description of changes:*

Found a bug during performance benchmarking where there were duplicates of all scheduler airflow processes running on a scheduler container (so 1 container had 2 schedulers running). 

This PR fixes this issue and also fixes issue where worker and webserver processes are not treated as essential (meaning failed conditions would not cause container to exit).

Tested deploying image locally and deploying to mwaa dev environment.

### DIFF
--- a/images/airflow/2.9.2/python/mwaa/entrypoint.py
+++ b/images/airflow/2.9.2/python/mwaa/entrypoint.py
@@ -414,6 +414,7 @@ def create_airflow_subprocess(
             "friendly_name": friendly_name,
             "conditions": conditions,
             "on_sigterm": on_sigterm,
+            "is_essential": True
     }
     if sigterm_patience_interval is not None:
         kwargs['sigterm_patience_interval'] = sigterm_patience_interval
@@ -501,6 +502,20 @@ def _get_sidecar_health_port():
         )
     except:
         return SIDECAR_DEFAULT_HEALTH_PORT
+
+
+def _create_airflow_webserver_subprocesses(environ: Dict[str, str]):
+    return [
+        create_airflow_subprocess(
+            ["webserver"],
+            environ=environ,
+            logger_name=WEBSERVER_LOGGER_NAME,
+            friendly_name="webserver",
+            conditions=[
+                AirflowDbReachableCondition(airflow_component="webserver"),
+            ],
+        ),
+    ]
 
 
 def _create_airflow_worker_subprocesses(environ: Dict[str, str], sigterm_patience_interval: timedelta | None = None):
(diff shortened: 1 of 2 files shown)


images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -602,13 +602,18 @@ def run_airflow_command(cmd: str, environ: Dict[str, str]):
...
-            run_subprocesses(_create_airflow_worker_subprocesses(environ))
+            run_subprocesses(
+                subprocesses=[],
+                essential_subprocesses=_create_airflow_worker_subprocesses(environ))
         case "webserver":
             run_subprocesses(
-                [
+                subprocesses=[],
+                essential_subprocesses=[
                     create_airflow_subprocess(
```
ms964107, line 617: nit: should we also modulize this like the scheduler/worker above?

ms964107 (approved):

kuyperse (approved):

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -602,13 +602,18 @@ def run_airflow_command(cmd: str, environ: Dict[str, str]):
             subprocesses = _create_airflow_scheduler_subprocesses(environ, conditions)
             # Schedulers, triggers, and DAG processors are all essential processes and
             # if any fails, we want to exit the container and let it restart.
-            run_subprocesses(subprocesses, essential_subprocesses=subprocesses)
+            run_subprocesses(
+                subprocesses=[],
```
dhegberg, line 606: There is the health check that looks for running AIrflow processes, how was this presenting itself in that loop?

Should that loop verify that only 1 Airflow process is present?

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -602,13 +602,18 @@ def run_airflow_command(cmd: str, environ: Dict[str, str]):
             subprocesses = _create_airflow_scheduler_subprocesses(environ, conditions)
             # Schedulers, triggers, and DAG processors are all essential processes and
             # if any fails, we want to exit the container and let it restart.
-            run_subprocesses(subprocesses, essential_subprocesses=subprocesses)
+            run_subprocesses(
```
ashishgo-aws, line 605: In the `run_subprocesses` implementation, the docs mention that
```
:param subprocesses: A list of Subprocess objects to run in parallel.
    :param essential_subprocesses: A sub-list of the processes that that must continue
      running, otherwise all sub-processes will be terminated.
```

But then, later this behavior got altered when this line was introduced within run_subprocesses
```
all_processes = subprocesses + essential_subprocesses
```
essential_subprocesses was supposed to be a sub list of subprocesses. 

I think a cleaner approach would be to make create_airflow_subprocess method accept an additional parameter named `essential` which can be used by the run_subprocesses method to check if a failed condition should cause the container to exit or not instead of passing two separate array of subprocesses.

ashishgo-aws, line 605: In the `run_subprocesses` implementation, the docs mention that
```
:param subprocesses: A list of Subprocess objects to run in parallel.
    :param essential_subprocesses: A sub-list of the processes that that must continue
      running, otherwise all sub-processes will be terminated.
```

But then, later this behavior got altered when this line was introduced within run_subprocesses
```
all_processes = subprocesses + essential_subprocesses
```
essential_subprocesses was supposed to be a sub list of subprocesses. 

I think a cleaner approach would be to make create_airflow_subprocess method accept an additional parameter named `essential` which can be used by the run_subprocesses method to check if a failed condition should cause the container to exit or not instead of passing two separate array of subprocesses.

ashishgo-aws (changes requested):

images/airflow/2.9.2/python/mwaa/entrypoint.py
Kytha: The healthcheck that does this is the ECS healthcheck. If failed, this healthcheck will bounce the containers. So I am not sure if this would be the best mechanism to detect an issue like this, since bouncing the containers wouldn't fix the issue and just create further interruptions / issues. 

I think we should invest in some "invasive" testing in our image build pipeline against a running container to catch issues like these.

dhegberg (approved):

kuyperse (approved):

ms964107 (approved):

ashishgo-aws (approved):

PR #134, opened 2024-08-27
Refactor logging to be async

Kytha:
*Description of changes:*

Performance testing MWAA airflow images revealed flaws with how logging is plumbed to CloudWatch. 

Currently, when airflow publishes log statements it is sent to an os pipe to be polled and read by the entrypoint process and then forwarded to cloudwatch. When this pipe is full, airflow must halt until logs are read before continuing execution. This means airflow execution is upper-bounded by how fast the entrypoint can read logs. 

The entrypoint has many other priorities in this critical loop (heartbeat healthcheck, db healthcheck, polling other subprocess logs, etc) and it has be measured that airflow scheduler is significantly throttled by this mechanism when log volume is large.

This PR aims to fix this by separating log polling into its own thread. This allows for faster log reading that is decoupled from the execution of the entrypoint process loop. Below is "before" and "after" profiler flamegraph results of a mwaa scheduler over a 1-minute period of execution under heavy scheduling and log load.

### Before
Screenshot 2024-08-27 at 1 43 08 AM

### After
Screenshot 2024-08-27 at 1 40 26 AM

In the before, we see that waiting for the log message to be read by the entry point is all the scheduler was able to do in the entire 1-minute interval. However with async logging, scheduler is able to complete multiple loops.

### DIFF
--- a/images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py
+++ b/images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py
@@ -35,7 +35,7 @@ _SIGTERM_DEFAULT_PATIENCE_INTERVAL = timedelta(seconds=90)
 
 # The time for which the log reader will sleep when subprocess is still running
 # But last log read was empty.
-_SUBPROCESS_LOG_POLL_IDLE_SLEEP_INTERVAL = timedelta(milliseconds=50)
+_SUBPROCESS_LOG_POLL_IDLE_SLEEP_INTERVAL = timedelta(milliseconds=100)
 
 
 module_logger = logging.getLogger(__name__)
@@ -223,12 +223,11 @@ class Subprocess:
         Execute a single iteration of the execution loop.
 
         The execution loop is a loop that continuously monitors the status of the
-        process and captures logs if any. This method executes a single iteration and
-        exit, expecting the caller to repeatedly call this method until it returns
-        `False`, meaning that the process has exited and there are no more logs to
-        ingest.
+        process. This method executes a single iteration and exits, expecting the
+        caller to repeatedly call this method until it returns
+        `False`, meaning that the process has exited.
 
-        :return: True if the process is still running and/or there are more logs to read.
+        :return: True if the process is still running.
         """
         if not self.process:
             # Process is not running anymore.
@@ -244,9 +243,7 @@ class Subprocess:
         if self.process_status == ProcessStatus.FINISHED:
             # We are done; call shutdown to ensure that we free all resources.
             self.shutdown()
-        elif self.process_status in [
-            ProcessStatus.RUNNING
-        ]:
+        elif self.process_status == ProcessStatus.RUNNING:
             # The process is still running, so we need to check conditions.
             failed_conditions = self._check_process_conditions()
 
@@ -295,9 +292,7 @@ class Subprocess:
             return
         self.is_shut_down = True
 
-        if self.process and self.process_status in [
-            ProcessStatus.RUNNING
-        ]:
+        if self.process and self.process_status == ProcessStatus.RUNNING:
             self._shutdown_python_subprocess(self.process)
         self.process = None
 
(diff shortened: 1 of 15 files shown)


dhegberg:
On first inspection the change makes sense.

What is the cause of the long schedule_run_dag in the second image?

images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py
```diff
@@ -32,6 +33,10 @@
 # from us before we forcefully terminate the process with a SIGKILL.
 _SIGTERM_DEFAULT_PATIENCE_INTERVAL = timedelta(seconds=90)
 
+# The time for which the log reader will sleep when subprocess is still running
+# But last log read was empty.
+_SUBPROCESS_LOG_POLL_IDLE_SLEEP_INTERVAL = timedelta(milliseconds=50)
```
ms964107, line 38: Is it too robust given that before we sleep roughly 1 second between each loop?

images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py
```diff
@@ -187,6 +195,29 @@ def _check_process_conditions(self) -> List[ProcessConditionResponse]:
...
+                    time.sleep(_SUBPROCESS_LOG_POLL_IDLE_SLEEP_INTERVAL.total_seconds())
+            else:
+                self.process_logger.info(line.decode("utf-8"))
+    
+    def _get_subprocess_status(self, process: Popen[Any]):
+        return ProcessStatus.RUNNING if process.poll() is None else ProcessStatus.FINISHED
+
     def execution_loop_iter(self):
         """
         Execute a single iteration of the execution loop.
```
ms964107, line 223: Update the description of this function

images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py
```diff
@@ -203,23 +234,18 @@ def execution_loop_iter(self):
...
-        )
-
-        if process_completed_and_logs_processed:
+        if self.process_status == ProcessStatus.FINISHED:
             # We are done; call shutdown to ensure that we free all resources.
             self.shutdown()
         elif self.process_status in [
-            ProcessStatus.RUNNING_WITH_NO_LOG_READ,
-            ProcessStatus.RUNNING_WITH_LOG_READ,
+            ProcessStatus.RUNNING
```
ms964107, line 248: `self.process_status == ProcessStatus.RUNNING`?

Kytha:
> On first inspection the change makes sense.
> 
> What is the cause of the long schedule_run_dag in the second image?

This call was during scheduler process of making [scheduling decision for a given dag run](https://github.com/apache/airflow/blob/2.9.2/airflow/jobs/scheduler_job_runner.py#L1469). If the given dag run has finished, then the [db session is flushed](https://github.com/apache/airflow/blob/2.9.2/airflow/models/dagrun.py#L918) to sync session with db. This is what took so long. So it is the case that during the 1-minute period I polled, a task happened to finish and triggered this code. 

It does take a suspiciously long time to do this, and may be an area of optimization in the future, but I don't believe it is related to any logging changes here.

images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py
Kytha: Yeah, this is something to considered. The issue here is that even a small environment can be a noisy logger. The goal here would be to poll faster than logs are produced by airflow. Which means we are trying to track an unknown quantity, since this will vary component to component, environment to environment, and also vary over time for same component / environment. 

Since the sleep is only when buffer is empty, maybe an increase to 100ms is warranted to be more cautious.

ashishgo-aws (approved):
Overall, changes looks good to me. After addressing the current comments, we should be good to go.

kuyperse (approved):

ashishgo-aws (approved):

dhegberg (approved):

PR #137, opened 2024-09-03
Adding signal processing to the task monitor condition

ashishgo-aws:
*Description of changes:*

The task monitor needs to process the following 4 MWAA signals for the graceful update project:
1. Termination signal: Graceful termination of the workers when the environment is going through a graceful update
2. Resume signal: Reverting the state of graceful termination and resume work when the environment is going through a rollback after attempting a graceful update
3. Kill signal: Shutting down the worker without waiting for the current Airflow tasks to finish when the environment is going through a forced update
4. Activation signal: Starting consumption of work from the queue after termination protection has enabled on the corresponding Fargate task

The processing is gated behind certain environment variables which are either absent or marked as false for an environment which does not have graceful updates enabled.


images/airflow/2.9.2/BillOfMaterials/2.9.2-dev/Python-Packages-BOM.txt
```diff
@@ -5611,10 +5611,10 @@ Copyright 2013-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 /usr/local/airflow/.local/lib/python3.11/site-packages/boto3-1.34.106.dist-info/NOTICE
 
 boto3-stubs
-1.35.6
```
ashishgo-aws, line 5614: In my change, the version is getting downgraded. I believe I am missing a step which ensures that the latest safe version is picked.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -70,6 +72,14 @@
 # consumption, we reset the worker to non-idle state and backoff from checking for
 # idleness for a minute.
 IDLENESS_RESET_BACKOFF_PERIOD = timedelta(minutes=1)
+# In case of issues, signals can arrive late and out of order. So, we scan all unprocessed signals from the last one hour.
```
dhegberg, line 75: Can you make all the comments generic related to the timedelta?  They have a tendency to be missed when someone adjusts the duration and then it will be confusing.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -136,14 +147,16 @@ def _create_shared_mem_celery_state():
 # will use to signal the toggle of a flag which tells the Celery SQS channel to
 # pause/unpause further consumption of available SQS messages. It is maintained by the
 # worker monitor.
-def _create_shared_mem_work_consumption_block():
+# When MWAA signal handling is enabled, the consumption will be turned off by default and
```
dhegberg, line 150: Maybe: `the consumption` -> `Airflow Task consumption`

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -136,14 +147,16 @@ def _create_shared_mem_celery_state():
...
+# it will be enabled only when the activation signal has been received by the worker.
+def _create_shared_mem_work_consumption_block(mwaa_signal_handling_enabled: bool):
     celery_work_consumption_block_name = (
         f'celery_work_consumption_{os.environ.get("AIRFLOW_ENV_ID", "")}'
     )
     celery_work_consumption_flag_block = shared_memory.SharedMemory(
```
dhegberg, line 156: Does the SharedMemory need to be unlinked?

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -273,15 +286,48 @@ def _cleanup_undead_process(process_id: int):
...
+    signal_filenames = os.listdir(MWAA_SIGNALS_DIRECTORY) if os.path.exists(MWAA_SIGNALS_DIRECTORY) else []
+    sorted_filenames = sorted(signal_filenames)
+    for signal_filename in sorted_filenames:
+        # Signals file follow the naming convention "<createTimestamp>_<signalId>.json"
+        file_timestamp = signal_filename.split("_")[0]
+        # In case of issues, signals can arrive late and out of order. So, we scan all unprocessed signals in a search time range.
+        # Only a handful of signals are expected for each worker, so this repeated processing should be very light.
+        if file_timestamp.isdigit() and int(file_timestamp) > signal_search_start_timestamp:
+            signal_file_path = os.path.join(MWAA_SIGNALS_DIRECTORY, signal_filename)
+            with open(signal_file_path, "r") as file_data:
```
dhegberg, line 300: What user is this process running as?  The files will be written as read-only to anything outside of the sidecar, correct?

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -273,15 +286,48 @@ def _cleanup_undead_process(process_id: int):
...
+                signal_data = json.load(file_data)
```
dhegberg, line 301: Just in case we have issues will writing malformed JSON, do we want to catch load exceptions and continue gracefully?

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -273,15 +286,48 @@ def _cleanup_undead_process(process_id: int):
...
+                if not signal_data["processed"]:
+                    return signal_file_path, signal_data
+    return None, None
+
+
+def _marked_signal_as_processed(signal_filepath, signal_data):
+    signal_data["processed"] = True
```
dhegberg, line 308: Following from above, if this user has write access do we have the risk of customers modifying/deleting the signals?  What is the behavior in that case?

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -362,6 +445,65 @@ def _get_current_task_count(self):
                 current_task_count += 1
         return current_task_count
 
+    def process_next_signal(self):
+        """
+        This method is used to process any signals sent by MWAA. This method processes the first signal it finds
+        in the chronological order of the available unprocessed signals.
+        """
+        if not self.mwaa_signal_handling_enabled:
+            logger.info("Signal handling is not enabled for this worker.")
```
dhegberg, line 454: Which log stream will this show up in?

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -70,6 +72,14 @@
 # consumption, we reset the worker to non-idle state and backoff from checking for
 # idleness for a minute.
 IDLENESS_RESET_BACKOFF_PERIOD = timedelta(minutes=1)
+# In case of issues, signals can arrive late and out of order. So, we scan all unprocessed signals from the last one hour.
+# Only a handful of signals are expected for each worker, so this repeated processing should be very light.
+SIGNAL_SEARCH_TIME_RANGE = timedelta(hours=1)
+# If a worker activation signal has not been received in the first 10 minute. Then, we give up on waiting anymore for the signal.
+ACTIVATION_WAIT_TIME_LIMIT = timedelta(minutes=10)
```
dhegberg, line 79: Should these all be defaults that can be overridden with Environment Variables?

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -376,8 +518,17 @@ def pause_task_consumption(self):
...
         self.celery_work_consumption_block.buf[0] = 1
+        if was_consumption_unpaused:
+            # When we toggle the consumption to paused state, we wait 10 seconds in order
+            # for any in-flight messages in the SQS broker layer to be processed and
+            # corresponding Airflow task instance to be created. Once that is done, we can
+            # start gracefully shutting down the worker. Without this, the SQS broker may
+            # consume messages from the queue, terminate before creating the corresponding
+            # Airflow task instance and abandon SQS messages in-flight.
+            logger.info("Pausing task consumption.")
+            time.sleep(10)
```
dhegberg, line 531: Why 10 seconds?

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: Not sure if I follow.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: We can do that. However, there are a lot of these variables and I am not sure of the value of adding the additional logic. If the variables are introduced, in order to change the behavior, the value of the variable will have to be changed which would still require an environment update.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: Sure

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: The code comment in the `close` method explains this. I am pasting it below for reference:
```
This will be called when the worker has been terminated. We close these blocks
        in order to avoid receiving a warning message in the logs. We are not calling
        unlink here intentionally because both the task monitor and the Celery SQS
        channel references these blocks and when the worker is shutting down, we do not
        control which process will be killed first. Calling unlink may result in a
        warning message showing up in the customer side logs causing unnecessary
        confusion.
```

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: In the serviceEni container implementation, the permissions are being explicitly changed for each file created. The Airflow container is able to read and write to the created files.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: Yes, we should.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: These files are created by the serviceEni container. The serviceEni container will look into just the data relevant to the corresponding Fargate task which means that these files will be specific to the worker processing them. 

If the customer for some reason modify these files, then they can make the task monitor make wrong decisions regarding when the worker should terminate or not. But none of the service side state is affected by it and we will have timeouts in place at the service-side to force correct this.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: All logger.info statements target service side logs stream here.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: I was thinking of keeping a 10 second value for the case where we resume the consumption. My thought process was that these 10 seconds will allow enough time for the worker to pick any work if available from the queue. This will cause the worker to become busy after its consumption has been resumed. And from there, I thought of keeping things consistent between the resume and the pause cases.

But thinking about this again, we have been using 5 seconds for the pause case with improved autoscaling with no issues. Perhaps, I should change them both to 5 seconds.

images/airflow/2.9.2/BillOfMaterials/2.9.2-dev/Python-Packages-BOM.txt
ashishgo-aws: This will be resolved in the next commit.

Kytha (changes requested):

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -55,21 +57,34 @@
...
 CLEANUP_ABANDONED_RESOURCES_DELAY_PERIOD = timedelta(minutes=1)
-# If in the improbable case of the worker picking up new tasks after having paused its
-# consumption, we reset the worker to non-idle state and backoff from checking for
-# idleness for a minute.
+# If in the improbable case of the worker picking up new tasks after having paused its consumption, we reset the worker to non-idle
+# state and backoff from checking for idleness for a certain threshold.
 IDLENESS_RESET_BACKOFF_PERIOD = timedelta(minutes=1)
+# In case of issues, signals can arrive late and out of order. So, we scan all unprocessed signals within a time range.
+# Only a handful of signals are expected for each worker, so this repeated processing should be very light.
+SIGNAL_SEARCH_TIME_RANGE = timedelta(hours=1)
```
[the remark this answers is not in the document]
ashishgo-aws: I think I understand. Here is my rationale behind the current state of things: the main and sidecar container each have their own set of responsibility and each is given 1 hour to ensure that they are able to finish their part. The 1 hour limit starts from the point in time of creation of their corresponding file and not the creation of the signal, so issues in sidecar container does not reduce the time given to the main container to finish its processing. This means that this config does not need to be in line with the sidecar config and it can be changed independently if needed.

images/airflow/2.9.2/python/mwaa/subprocess/conditions.py
```diff
@@ -496,34 +516,43 @@ def _check(self, process_status: ProcessStatus) -> ProcessConditionResponse:
...
+                logger.info("Worker process is either NOT idle or has not been marked for termination. No action is needed.")
         else:
             logger.info(
                 f"Worker process finished (status is {process_status.name}). "
-                "No need to check for idleness anymore."
+                "No need to monitor tasks anymore."
             )
 
-        # The AutoScalingCondition always succeeds, unless the worker is idle in which
+        # The TaskMonitoringCondition always succeeds, unless the worker is idle in which
```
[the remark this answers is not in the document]
ashishgo-aws: I will update this in the next commit after clubbing the change with other feedback that I might receive.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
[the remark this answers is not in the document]
Kytha: Ah, I do think that the sidecar and main container can operate independent of this config, so I think this is fine. I was confusing myself between the EFS and EBS volumes.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -273,25 +291,62 @@ def _cleanup_undead_process(process_id: int):
     )
 
 
+def _get_next_unprocessed_signal() -> (str, dict):
```
Kytha, line 294: What happens if a signal is sent, then 30 minutes later container exits due to some unrelated issue and is replaced. Since sidecar syncs anything in past hour, will this new container re-process this old signal. Do we need a mechanism preventing signals from being processed that predate container launch?

Kytha (changes requested):

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: When the container exits, the corresponding Fargate task will be terminated and replaced with a new one with its own Fargate task ID and a new EBS volume. There is no data from the old Fargate task/container in the EBS volume that could get processed by the new containers. And when the serviceEni container lookup for data in the EFS volume, it only takes data corresponding to the current Fargate task, so again any old data for the previous task is not picked up.

dhegberg (approved):

dhegberg (approved):

millin:
Doesn't this need to be ported to the 2.10.1?

ashishgo-aws:
> Doesn't this need to be ported to the 2.10.1?

Yes. PR for that: https://github.com/aws/amazon-mwaa-docker-images/pull/151

Please note that this is for a feature under development and is currently disabled. So, functionally speaking, 2.9 and 2.10 images are the same.

PR #135, opened 2024-08-31
Update references to subprocess status

Kytha:
*Issue #, if available:* N/A

Fixing references to old subprocess status in AutoScalingCondition. The RUNNING_WITH_LOG_READ and RUNNING_WITH_NO_LOG_READ no longer exist and have been replaced by the RUNNING status.

PR #139, opened 2024-09-05
Add 2.10.1 Airflow Docker image

michaelfu1029:
*Issue #, if available:*

*Description of changes:*
Adding the folder and code for the Airflow 2.10.1 image. Copied the 2.9.2 image and updated the version and constraints. Built and ran locally with the run.sh script.


mayushko26 (dismissed):
Approving as it's a no-impact change, and will be iterated upon in subsequent PR's. However, this change is nearly impossible for a reviewer to confidently review due to the volume of files, so we should give some thought into how we can improve this initial image creation process. Perhaps we could split it up into two PR's: one to copy files into a new x.y.z folder; and a second to make the version-specific changes. This would show the effective diff in the second PR, and draw more attention to the substantive changes made between versions.

Update: The PR now includes two commits - one for copying the 2.9.2 folder into a new 2.10.0 folder, and another for making the version-specific changes. Reviewers can focus on the second commit.

images/airflow/2.10.0/python/mwaa/subprocess/subprocess.py
```diff
@@ -188,7 +188,7 @@ def start(
                 exc_info=sys.exc_info(),
             )
 
-    @throttle(seconds=60, instance_level_throttling=True) # avoid excessive calls to process conditions
```
mayushko26: It looks like a few of the recent changes (https://github.com/aws/amazon-mwaa-docker-images/pull/136, https://github.com/aws/amazon-mwaa-docker-images/pull/135) haven't been included here. Probably requires a rebase.

michaelfu1029: Fixed this

mayushko26 (dismissed):

vishalvijay18 (dismissed):

images/airflow/2.10.1/bin/airflow-user/post-startup-script-verification
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 echo "Executing post startup script verification"
 
-EXPECTED_AIRFLOW_VERSION="2.10.0"
-EXPECTED_WATCHTOWER_VERSION="3.2.0"
+EXPECTED_AIRFLOW_VERSION="2.10.1"
+EXPECTED_WATCHTOWER_VERSION="3.3.1"
```
michaelfu1029, line 5: Bumping this as it is changed in the constraints for this version. We will test logging to make sure it doesn't break. Rafid revamped logging for 2.9 so we expect it to go smoothly

mayushko26 (approved):

jessefrn1 (approved):

vishalvijay18 (approved):

PR #150, opened 2024-09-27
Letting each condition control its frequency

ashishgo-aws:
Using throttling, each condition was being checked once per minute. But a condition may need to perform its check at a different pace like for example the TaskMonitor condition. This change applies throttling at individual condition instead of applying it generically at the subprocess level. The TaskMonitor condition will not be executed once per 10 seconds instead of one minute. This will allow for faster evaluation of worker idleness which would help in scaling down the worker fleet a lot more effectively.

### DIFF
--- a/images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
+++ b/images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
@@ -46,7 +46,7 @@ EOF_TOKEN = "EOF_TOKEN"
 # blocks without running out of space.
 BUFFER_SIZE_PER_TASK = 2500
 CELERY_WORKER_TASK_LIMIT = int(
-    os.environ.get("AIRFLOW__CELERY__WORKER_AUTOSCALE", "20,20").split(",")[0]
+    os.environ.get("AIRFLOW__CELERY__WORKER_AUTOSCALE", "80,80").split(",")[0]
 )
 CELERY_TASKS_BUFFER_SIZE = CELERY_WORKER_TASK_LIMIT * BUFFER_SIZE_PER_TASK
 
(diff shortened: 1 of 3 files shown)


images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -46,7 +46,7 @@
 # blocks without running out of space.
 BUFFER_SIZE_PER_TASK = 2500
 CELERY_WORKER_TASK_LIMIT = int(
-    os.environ.get("AIRFLOW__CELERY__WORKER_AUTOSCALE", "20,20").split(",")[0]
+    os.environ.get("AIRFLOW__CELERY__WORKER_AUTOSCALE", "80,80").split(",")[0]
```
dhegberg, line 49: How is this related to the change?

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
```diff
@@ -622,6 +622,7 @@ def close(self):
             return
 
         logger.info("Closing task monitor...")
+        self.pause_task_consumption()
```
dhegberg, line 625: Like above.

images/airflow/2.9.2/python/mwaa/celery/task_monitor.py
ashishgo-aws: It's actually not. These are small improvements that I found while working on the porting code for older and newer versions. Original plan was to add these as a separate PR but this PR was sitting in stale state for quite some time and I merged those changes to this PR to save time. I can remove them if needed.

kuyperse (changes requested):
These changes also need to be ported to 2.10.1.

ashishgo-aws:
> These changes also need to be ported to 2.10.1.

I believe I have covered these in https://github.com/aws/amazon-mwaa-docker-images/pull/151. Is there anything you saw that is missing in that PR?

kuyperse (approved):

kuyperse:
> > These changes also need to be ported to 2.10.1.
> 
> I believe I have covered these in #151. Is there anything you saw that is missing in that PR?
I didn't check the other PRs to see if the changes were ported there. Nothing seems to be missing. LGTM.

dhegberg (approved):

PR #168, opened 2024-10-30
Update pool size if environment was created during problematic timeframe

michaelfu1029:
*Issue #, if available:*

*Description of changes:*
2.9.2 environments created after July 8(launch) and before Sept 6(fix deployed) have default_pool size 128 so we update this to 10000 (the value used for previous versions)

### DIFF
--- a/images/airflow/2.9.2/python/mwaa/entrypoint.py
+++ b/images/airflow/2.9.2/python/mwaa/entrypoint.py
@@ -126,7 +126,7 @@ async def increase_pool_size_if_default_size(environ: dict[str, str]):
     :param environ: A dictionary containing the environment variables.
     """
     created_at = os.environ.get("MWAA__CORE__CREATED_AT")
-    problemaic_pool_size = 128
+    problematic_pool_size = 128
 
 
     if created_at:
@@ -144,7 +144,7 @@ async def increase_pool_size_if_default_size(environ: dict[str, str]):
                                env=environ, stdout_logging_method=lambda output : command_output.append(output))
 
             # Increasing the pool size if it is the default size
-            if len(command_output) == 1 and int(command_output[0]) == DEFAULT_POOL_SIZE:
+            if len(command_output) == 1 and int(command_output[0]) == problematic_pool_size:
                 logger.info("Setting default_pool size to 10000.")
                 await run_command("airflow pools set default_pool 10000 default", env=environ)
 
--- a/images/airflow/2.9.2/python/mwaa/entrypoint.py
+++ b/images/airflow/2.9.2/python/mwaa/entrypoint.py
@@ -66,7 +66,7 @@ from mwaa.subprocess.conditions import (
 from mwaa.subprocess.subprocess import Subprocess, run_subprocesses
 from mwaa.utils.cmd import run_command
 from mwaa.utils.dblock import with_db_lock
-
+from mwaa.utils.statsd import get_statsd
 
 # Usually, we pass the `__name__` variable instead as that defaults to the
 # module path, i.e. `mwaa.entrypoint` in this case. However, since this is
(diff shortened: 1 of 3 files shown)


images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -116,6 +117,37 @@ async def airflow_db_init(environ: dict[str, str]):
...
+        created_date = datetime.strptime(created_at, date_format)
+        # Has a little buffer from when 2.9.2 was released and when the fix was fully deployed
+        issue_beginning = datetime(2024, 7, 8)
+        issue_resolution = datetime(2024, 9, 6)
+
+        if created_date > issue_beginning and created_date < issue_resolution:
+            command_output = []
+
+            # Get the current default_pool size
+            await run_command("airflow pools get default_pool | grep default_pool | awk '{print $3}'",
```
mayushko26, line 143: Is there a reason we can't use Pool.get_pool("default_pool") instead? Reference: https://github.com/apache/airflow/blob/2.9.2/airflow/models/pool.py#L66.

I don't doubt this works, but it's somewhat opaque as-is.

michaelfu1029: I remember trying to use this but I think there was some issues that came up with setting up the session since we are not running it as a dag.

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -116,6 +115,39 @@ async def airflow_db_init(environ: dict[str, str]):
...
+    Update the default pool size
+
+    Fix a regression where some 2.9.2 environments were created with the default 128 default
+    pool size. This function checks if the environment was created during the problematic
+    timeframe and update the size if it has not been updated by the customer.
+
+    :param environ: A dictionary containing the environment variables.
+    """
+    created_at = os.environ.get("MWAA__CORE__CREATED_AT")
+    problemaic_pool_size = 128
```
mayushko26, line 129: nit: spelling

images/airflow/2.9.2/python/mwaa/entrypoint.py
mayushko26: I see, makes sense. If you've time boxed it and could not get it to work with Pools, I think it's okay to leave it like this.

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -116,6 +115,39 @@ async def airflow_db_init(environ: dict[str, str]):
...
+
+        if created_date > issue_beginning and created_date < issue_resolution:
+            command_output = []
+
+            # Get the current default_pool size
+            await run_command("airflow pools get default_pool | grep default_pool | awk '{print $3}'",
+                               env=environ, stdout_logging_method=lambda output : command_output.append(output))
+
+            # Increasing the pool size if it is the default size
+            if len(command_output) == 1 and int(command_output[0]) == DEFAULT_POOL_SIZE:
```
mayushko26, line 147: DEFAULT_POOL_SIZE does not exist with this revision, I believe. Should this be problematic_pool_size instead?

mayushko26 (approved):

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -116,6 +115,39 @@ async def airflow_db_init(environ: dict[str, str]):
...
+            command_output = []
+
+            # Get the current default_pool size
+            await run_command("airflow pools get default_pool | grep default_pool | awk '{print $3}'",
+                               env=environ, stdout_logging_method=lambda output : command_output.append(output))
+
+            # Increasing the pool size if it is the default size
+            if len(command_output) == 1 and int(command_output[0]) == problematic_pool_size:
+                logger.info("Setting default_pool size to 10000.")
+                await run_command("airflow pools set default_pool 10000 default", env=environ)
```
Kytha, line 149: It may be nice to emit some emf metric here so we can track how many environments this has fixed

Kytha (changes requested):

images/airflow/2.9.2/python/mwaa/entrypoint.py
```diff
@@ -116,6 +115,39 @@ async def airflow_db_init(environ: dict[str, str]):
...
+    Update the default pool size
+
+    Fix a regression where some 2.9.2 environments were created with the default 128 default
+    pool size. This function checks if the environment was created during the problematic
+    timeframe and update the size if it has not been updated by the customer.
+
+    :param environ: A dictionary containing the environment variables.
+    """
+    created_at = os.environ.get("MWAA__CORE__CREATED_AT")
+    problematic_pool_size = 128
```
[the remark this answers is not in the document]
michaelfu1029: I had this initially as that but it was confusing having the bad default as just DEFAULT_POOL_SIZE.

mayushko26 (approved):

Kytha (approved):

PR #183, opened 2024-11-22
Update MariaDB distro and checksum

michaelfu1029:
*Issue #, if available:*

*Description of changes:*
Updating MariaDB distro to use 11.4

Tested on both x86 and arm


dhegberg:
Was this tested as a local docker cluster or in the MWAA environment?

This change from the 11.3 notes shouldn't have an impact an impact (we're enforcing SSL) but I'd like to make sure it works.  This should be obvious in our pipeline tests too though, just keep an eye out during deployment.

```
*[SSL](https://mariadb.com/kb/en/data-in-transit-encryption/) is now enabled in the server by default. No configuration necessary, if no server certificate was provided a self-signed certificate will be automatically generated by the server. See [Mission Impossible: Zero-Configuration SSL](https://mariadb.org/mission-impossible-zero-configuration-ssl/) on mariadb.org ([MDEV-31856](https://jira.mariadb.org/browse/MDEV-31856)).
*Clients now can validate self-signed server certificates if the [mysql_native_password](https://mariadb.com/kb/en/authentication-plugin-mysql_native_password/) or [ed25519](https://mariadb.com/kb/en/authentication-plugin-ed25519/) authentication is used and account password is not empty ([MDEV-31855](https://jira.mariadb.org/browse/MDEV-31855)).
*Clients now require SSL and have [--ssl-verify-server-cert](https://mariadb.com/kb/en/mariadb-command-line-client/#-ssl-verify-server-cert) enabled by default ([MDEV-31857](https://jira.mariadb.org/browse/MDEV-31857)).
```
https://mariadb.com/kb/en/mariadb-11-3-0-release-notes/

dhegberg (approved):

michaelfu1029:
It was tested locally running the `run.sh` script. I ran without the change and it gave the mariadb error but after the change it succeeded

Issue #196, opened 2024-11-28
Generate valid fernet keys for local deployment


Kytha:
## Overview

The [current fernet key](https://github.com/aws/amazon-mwaa-docker-images/blob/f71d082cac983385349f3e30b8a650d5f5c5d619/images/airflow/2.9.2/docker-compose.yaml#L23) in docker-compose file is deliberately invalid. This prevents the use of any feature relying on fernet encryption for local deployments.

To prevent hard-coding a valid fernet key (which risks general public using it in production), we should update run.sh to generate a valid fernet key, cache it locally, and pass it as an environment variable to the docker-compose file. So when run.sh is executed it either gets key from the cache, or if it doesn't exist, it generates a new one.

PR #207, opened 2025-01-03
Generate, cache and pass valid fernet keys for local deployment (#196)

kashyapkannan:
*Issue #, if available:* #196

*Description of changes:* 
- Modified run.sh to generate, cache and pass valid fernet key to the docker-compose file.
- Added a script to be able to use pip install in run.sh to temporarily install dependencies needed before the bootstrap steps, like in this use-case.
- Updated pip_install_check.py and run_all.py quality_check files' shebang for better portability.
- Updated lint_bash check to exclude .venv's generated scripts.

*Description of testing:* 
- Built and ran image locally with the run.sh script.
- Used log statements to verify the fernet key was valid and being passed correctly.

### DIFF
--- a/images/airflow/2.9.2/run.sh
+++ b/images/airflow/2.9.2/run.sh
@@ -10,6 +10,37 @@ else
     CONTAINER_RUNTIME="docker"
 fi
 
+# Generate valid Fernet key as json
+generate_fernet_key() {
+
+    # Install cryptography package quietly
+    chmod +x temporary-pip-install generate_fernet_key.py
+    ./temporary-pip-install cryptography >/dev/null 2>&1
+    
+    # Generate the key and format as JSON
+    KEY=$(python3 generate_fernet_key.py)
+    
+    # Uninstall cryptography package quietly
+    python3 -m pip uninstall -y cryptography cryptography-vectors &>/dev/null 2>&1
+    
+    echo "$KEY"
+}
+
+# Set up cache directory ; generate if it dosen't exist
+CACHE_DIR="${HOME}/.cache/mwaa-local"
+FERNET_KEY_FILE="${CACHE_DIR}/fernet.key"
+mkdir -p "${CACHE_DIR}"
+
+# Check if we have a cached Fernet key, if not generate and cache it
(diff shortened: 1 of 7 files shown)


images/airflow/2.9.2/run.sh
```diff
@@ -10,6 +10,37 @@ else
     CONTAINER_RUNTIME="docker"
 fi
 
+# Generate valid Fernet key as json
+generate_fernet_key() {
+
+    # Install cryptography package quietly
+    chmod +x temporary-pip-install
+    ./temporary-pip-install cryptography >/dev/null 2>&1
```
mayushko26, line 18: I'm wondering why we are using a separate utility for this, rather than a direct pip3 install command. Is this to get around linting?

images/airflow/2.9.2/run.sh
```diff
@@ -10,6 +10,37 @@ else
...
+    
+    # Generate the key and format as JSON
+    KEY=$(python3 -c 'from cryptography.fernet import Fernet; import json; key = Fernet.generate_key().decode(); print(json.dumps({"FernetKey": key}))')
```
mayushko26, line 21: Should this code live into a .py file instead of an inline string? For readability, linting, etc.

images/airflow/2.9.2/run.sh
kashyapkannan: Yes, since we can't directly use pip3 install ([refer](https://github.com/aws/amazon-mwaa-docker-images/blob/main/quality-checks/pip_install_check.py)) , and this use-case doesn't fit the 'safe-pip-config' utility as well, since we are using and uninstalling the package before the bootstrap steps that involve safe-pip-install. So I've created and used a utility for this use-case, and made it generic to support similar cases in the future.

images/airflow/2.9.2/run.sh
kashyapkannan: Yeah I think that will make it better, will implement this change

vishalvijay18 (approved):

mayushko26 (approved):

Issue #176, opened 2024-11-15
Use an init process as container entrypoint


Kytha:
## Overview

Our image runs the entry point script directly as the container entry point. This is against docker best practices because it omits the use of an init process. This will mean that no zombie processes will get properly reaped and will accumulate until system hits process limit. This is a problematic issue for a software like airflow, that dynamically creates subprocesses regularly. 

Instead, our image should follow [airflow](https://airflow.apache.org/docs/docker-stack/entrypoint.html#signal-propagation) and [docker best practices](https://docs.docker.com/engine/containers/multi-service_container/) and make use of a lightweight init process like [dumb-init](https://github.com/Yelp/dumb-init) or [tini](https://github.com/krallin/tini).

This issue may be the reason why we sometimes see issues related to delays in worker task shutdown, or schedulers stalling after long lifetimes. 

## Acceptance Criteria

* Integrate a lightweight init process to all airflow images

## Additional Info

Also see related discussion in github airflow issue: https://github.com/apache/airflow/discussions/25225

PR #209, opened 2025-01-07
Added tini as init process to all airflow images. (#176)

sudiptoghosh01:
Issue #, if available: [#176](https://github.com/aws/amazon-mwaa-docker-images/issues/176)

*Description of changes:*
- Updated `Dockerfile.base.j2` file to include necessary url components to download `tini`.
- Updated `Dockerfile.derivatives.j2`, set the container entrypoint to use `tini` as init system and launches MWAA's Python entrypoint module.
- Added `tini` installation and permissions setup in `001-init.sh` script.
- Made the above changes to all airflow images.

### UPDATE
Going forward to use docker's inbuilt init. Hence, enabling it in docker compose.

*Description of testing:*
- Built and ran image locally with the run.sh script.
- Checked if tini is running as PID#1 using `/proc/1/status`.

### DIFF
--- a/images/airflow/2.10.1/docker-compose-hybrid.yaml
+++ b/images/airflow/2.10.1/docker-compose-hybrid.yaml
@@ -3,6 +3,7 @@ version: "3.8"
 x-airflow-common: &airflow-common
   image: amazon-mwaa-docker-images/airflow:2.10.1-dev
   restart: always
+  init: true
   environment:
     # AWS credentials
     AWS_ACCESS_KEY_ID: ${AWS_ACCESS_KEY_ID}
--- a/images/airflow/2.10.1/docker-compose-resetdb.yaml
+++ b/images/airflow/2.10.1/docker-compose-resetdb.yaml
@@ -2,6 +2,7 @@ version: "3.8"
 
 x-airflow-common: &airflow-common
   image: amazon-mwaa-docker-images/airflow:2.10.1-dev
+  init: true
   environment:
     # AWS credentials
     AWS_ACCESS_KEY_ID: "FAKE_AWS_ACCESS_KEY_ID"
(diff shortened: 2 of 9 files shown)


images/airflow/2.10.1/Dockerfile.base.j2
```diff
@@ -31,6 +31,12 @@ ARG MARIADB_RPM_DEVEL_ARM_CHECKSUM=a49c1402378581ba2d63d97daa4a4c04
 ARG MARIADB_RPM_SHARED_PREFIX=MariaDB-shared-11.4.3-1.fc38
 ARG MARIADB_RPM_SHARED_X86_CHECKSUM=26a6c7fb021c09b85c21ec1878c1953f
 ARG MARIADB_RPM_SHARED_ARM_CHECKSUM=fd2ce395bf7d2cb959495bbd02ffec2e
+ARG TINI_DOWNLOAD_BASE_URL=https://github.com/krallin/tini/releases/download/v0.19.0/
```
vishalvijay18, line 34: Wondering if ECS fargate provides a direct way to run init process. I see mention of _initProcessEnabled_ in task definition parameters [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html).

sudiptoghosh01: `initProcessEnabled` is specific to ECS fargate. However, instances of these images can be deployed to any hosts. But looks like we can use `init` key in docker compose file that will automatically do the installation instead of doing that via shell. I will make the changes accordingly. Thanks.

Kytha: In generally, I like the idea of using tini directly in the image so that customers can pull our image and not worry about init process, plus you get guaranteed parity with how it is running on MWAA. 

However, my concern here is pulling from github during image build, at least for the MWAA use-case. I think we need a mechanism to be able to pull from some internal source / trusted vendor to avoid supply chain exposure. 

Othewise, maybe we opt to use the built-in tini with docker

Kytha (changes requested):

vishalvijay18 (approved):

Kytha (approved):

PR #213, opened 2025-01-27
Add test-requirements,test-startup-script, and resetdb script commands

michaelfu1029:
*Issue #, if available:*

*Description of changes:*
Added 3 new commands for testing similar to those from local runner. They can be ran like`./run.sh test-requirements`. Full instructions are in the readme. Purpose of them is to test dependencies without starting airflow.

Note: for resetdb, you have to manually uncomment a section in your versions entrypoint as we do not want this section enabled in production as having it execute on a customer's database would reset it


images/airflow/2.10.1/python/mwaa/entrypoint.py
```diff
@@ -709,6 +711,11 @@ async def main() -> None:
             **mwaa_essential_airflow_config,
         },
     )
+
+    if command == "test-startup-script":
```
vishalvijay18, line 715: Are we not adding full implementation as part of this commit?

michaelfu1029: The entrypoint already checks these different parts throughout the execution, if users want to test just the requirements for example, we can just end the execution early and not have airflow start. These statements are right after the check is done

images/airflow/2.10.1/run.sh
```diff
@@ -101,4 +105,10 @@ export MWAA__CORE__TASK_MONITORING_ENABLED
 export MWAA__CORE__TERMINATE_IF_IDLE
 export MWAA__CORE__MWAA_SIGNAL_HANDLING_ENABLED
 
-$CONTAINER_RUNTIME compose up
+if [ "$COMMAND" == "resetdb" ]; then
+    $CONTAINER_RUNTIME compose -f docker-compose-resetdb.yaml up --abort-on-container-exit
+elif [ "$COMMAND" == "test-requirements" ] || [ "$COMMAND" == "test-startup-script" ]; then
+    $CONTAINER_RUNTIME compose -f docker-compose-commands.yaml up "$COMMAND" --abort-on-container-exit
```
ashishgo-aws, line 111: Just a minor comment about the naming of the commands file. It is not following the current convention. Maybe we can have a separate file for each test commands like `docker-compose-resetdb.yaml`. Or, maybe we can rename this file as `docker-compose-test-commands.yaml` to suggest that this file is for test scripts only. Right now, it is a bit too generic.

images/airflow/2.10.1/python/mwaa/entrypoint.py
vishalvijay18: I see. In that case if we are running `test-requirements` then it will also execute startup script part. But that should be ok.

vishalvijay18 (dismissed):

images/airflow/2.10.1/run.sh
```diff
@@ -101,4 +105,10 @@ export MWAA__CORE__TASK_MONITORING_ENABLED
 export MWAA__CORE__TERMINATE_IF_IDLE
 export MWAA__CORE__MWAA_SIGNAL_HANDLING_ENABLED
 
-$CONTAINER_RUNTIME compose up
+if [ "$COMMAND" == "resetdb" ]; then
```
dhegberg, line 108: Is reset-db needed as a command?  In the testing scenario, doesn't `docker-compose down --volumes` result in the database container being removed.

Seems safer rather than introducing the reset-db that we don't plan on using in production.

mayushko26: Yeah, in that case we can update the README to describe that docker-compose command to accomplish the use case of reseting db.

michaelfu1029: Updated

images/airflow/2.10.1/run.sh
michaelfu1029: Changed name to this

mayushko26 (approved):

ashishgo-aws (approved):

PR #214, opened 2025-02-06
Use custom decoder over internal pip module for requirements scan

kuyperse:
*Issue #, if available:* N/A

*Description of changes:*
Pip 25.0 recently moved an internal decoding module used by MWAA airflow images to scan requirements files for constraints. The change breaks the constraint check in the images and results in Airflow constraints being forced always which overrides any custom constraints. This commit duplicates the previously used internal pip module and adds it to the utils folder. An additional log line is also added to log any errors that may occur in the future during the requirements scan.

---


Mercury2699:
Attaching the pull request from pip: https://github.com/pypa/pip/pull/12795/

Mercury2699 (approved):

dhegberg (approved):

PR #215, opened 2025-02-07
Adding fix for incorrect bug and refactoring entrypoint.py to fix circular dependency issue

Vraj2725:
*Issue #, if available:*
When using airflow versions 2.9.2 and above customer are unable to see error logs when they set their loglevel to `WARNING` or above. The root cause is that in subprocess.py, the line `self.process_logger.info(line.decode("utf-8"))` overrides the log level coming from airflow stream to INFO in [subprocess.py](https://github.com/aws/amazon-mwaa-docker-images/blob/main/images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py#L217).

*Description of changes:*
- Setting prcoess logger level to `DEBUG` to emit all logs and setting root logger to customer's log level to respect customer's log level.
- Refactoring `entrypoint.py` and breaking down its logic into different files so all files have single responsibility such as:
     - `user_requirements.py`: Installing user requirements.
     - `execute_command.py`: Logic for executing all the subprocesses
     - `setup_environment.py`: Running the startup script and setting up environment variables

### DIFF
--- a/images/airflow/2.10.1/python/mwaa/subprocess/subprocess.py
+++ b/images/airflow/2.10.1/python/mwaa/subprocess/subprocess.py
@@ -16,6 +16,7 @@ import atexit
 import fcntl
 import logging
 import os
+import re
 import signal
 import subprocess
 import sys
@@ -196,6 +197,42 @@ class Subprocess:
 
         return failed_conditions
 
+    def _parse_airflow_log(self, line: str):
+        log_pattern = re.compile(
+            r'\[(?P<timestamp>[^\]]+)\] '
+            r'\{(?P<component>[^}]+)\} '
+            r'\x1b?\[?\d{0,2}m?(?P<level>INFO|ERROR|WARNING|DEBUG)\x1b?\[?0?m? - '
+            r'(?P<message>.*)'
+        )
+        match = log_pattern.search(line)
+        if match:
+            return {
+                'timestamp': match.group("timestamp"),
+                'component': match.group("component"),
+                'level': match.group("level"),
+                'message': match.group("message")
+            }
+        return None
+
+    def _emit_logs(self, message: str):
+        parsed_log = self._parse_airflow_log(message)
+
+        if parsed_log:
+            level = parsed_log['level']
+            formatted_message = f"[{parsed_log['timestamp']}] {parsed_log['component']} {level} - {parsed_log['message']}"
+
+            if level == 'ERROR':
+                self.process_logger.error(formatted_message)
+            elif level == 'WARNING':
+                self.process_logger.warning(formatted_message)
+            elif level == 'DEBUG':
+                self.process_logger.debug(formatted_message)
+            else:  # INFO or undefined
+                self.process_logger.info(formatted_message)
+        else:
+            # Fallback to original behavior for non-matching logs
+            self.process_logger.info(message)
+
     def _read_subprocess_log_stream(self, process: Popen[Any]):
         """
         Poll process stdout and forward logs to subprocess logger
@@ -204,9 +241,11 @@ class Subprocess:
         for small duration to avoid wasted cpu resources.
         """
         stream = process.stdout
+
         while True:
             if not stream or stream.closed:
                 break
+
             line = stream.readline()
             if line == b"":
                 if process.poll() is not None:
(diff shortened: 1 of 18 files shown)


images/airflow/2.10.1/python/mwaa/subprocess/subprocess.py
```diff
@@ -196,6 +197,42 @@ def _check_process_conditions(self) -> List[ProcessConditionResponse]:
 
         return failed_conditions
 
+    def _parse_airflow_log(self, line: str):
+        log_pattern = re.compile(
```
dhegberg, line 201: Do we allow customers to adjust log formatting?
https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#logging

Wouldn't this break if the log format is different than the default?

mayushko26: Yeah they can configure their own format.

I would recommend a few things:
- Infer the log format based on the `AIRFLOW__LOGGING__LOG_FORMAT` configuration
- Regex is expensive, and we should test the case of high log throughput, and its effects on the health of the environment. There may be a more efficient way to retrieve the log level, based on the jinja template.
- Do we need to capture the timestamp, component, and message here? It seems like only the level is used to dictate behaviour here.

images/airflow/2.10.1/python/mwaa/subprocess/subprocess.py
```diff
@@ -196,6 +197,42 @@ def _check_process_conditions(self) -> List[ProcessConditionResponse]:
...
+                'message': match.group("message")
+            }
+        return None
+
+    def _emit_logs(self, message: str):
+        parsed_log = self._parse_airflow_log(message)
+
+        if parsed_log:
+            level = parsed_log['level']
+            formatted_message = f"[{parsed_log['timestamp']}] {parsed_log['component']} {level} - {parsed_log['message']}"
```
mayushko26, line 222: Some customers rely on a specific log format for their own monitoring and automations around their workflows. Ideally we shouldn't be reformatting the message that's coming from Airflow.

images/airflow/2.10.1/python/mwaa/subprocess/subprocess.py
mayushko26: Yes they can change it via the `AIRFLOW__LOGGING__LOG_FORMAT` configuration.

- I recommend using that configuration to infer the format
- Regex is expensive and log processing is a hot loop. We benchmark a case with high log throughput to ensure this doesn't lead to resource utilization spikes.
- Is it necessary to capture the timestamp, component, and message here? It seems like the log level is the only element used to dictate behaviour here.

Kytha: Agreed. I think their will be a significant performance hit by trying to parse each log line, in addition to concerns about brittle nature. 

Would a fast alternative be to not propagate the customer defined log level to the process_logger and instead set the process_logger level to DEBUG? This will allow for all logs emitted from airflow, regardless of level, to propagate to CloudWatch. Then, to respect the customer's log level we can configure the root logger in the config [here](https://github.com/aws/amazon-mwaa-docker-images/blob/main/images/airflow/2.9.2/python/mwaa/logging/config.py#L38) to be the customer's log level. 

This is similar to how we handled logging in previous versions. It also is more efficient than even the current state. Right now all airflow logs are still emitted to stdout and processed by our entrypoint script, even if customer increases level or outright disables logging. This has in the past diminished our ability to mitigate certain logging issues when necessary, since asking customer to disable logs will have no impact if bug / bottleneck is in our entrypoint prior to reaching watchtower.

Maybe there are shortcomings with this approach that necessitated the new logging behaviour in these images, but if not than this would be my recommendation.

Kytha (changes requested):

images/airflow/2.10.1/python/mwaa/subprocess/subprocess.py
Vraj2725: Thanks for the inputs. I will test the suggestion and raise a revision.

Vraj2725:
I implemented changes as per Kyle's suggestions. Have updated he description for the same.

images/airflow/2.10.3/python/mwaa/subprocess/subprocess.py
```diff
@@ -16,6 +16,7 @@
 import fcntl
 import logging
 import os
+import re
```
Kytha, line 19: import no longer needed

images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py
```diff
@@ -9,13 +9,15 @@
...
 from datetime import timedelta
+from os import environ
 from subprocess import Popen
 from types import FrameType, TracebackType
 from typing import Any, Callable, Dict, List, Optional
 import atexit
 import fcntl
 import logging
 import os
+import re
```
Kytha, line 20: Import not needed

Kytha (dismissed):

Kytha (approved):

kashyapkannan (approved):

PR #218, opened 2025-02-14
Reserialize dags on db migration

michaelfu1029:
*Issue #, if available:*

*Description of changes:*
In previous airflow versions, we had reserialize dags as true. Right now it does not so setting this config to True. Tested by having the run script go to this path and successfully execute the command with this new config.

### DIFF
--- a/images/airflow/2.10.3/python/mwaa/database/migrate.py
+++ b/images/airflow/2.10.3/python/mwaa/database/migrate.py
@@ -40,12 +40,22 @@ def _migrate_db():
         args = Namespace(migration_wait_timeout=1)
         airflow_db_command.check_migrations(args)
         logging.info("The database is already migrated.")
+        args = Namespace(
+            from_revision=None,
+            from_version=None,
+            reserialize_dags=True,
+            show_sql_only=None,
+            to_revision=None,
+            to_version=None,
+            use_migration_files=None,
+        )
+        airflow_db_command.migratedb(args)
     except TimeoutError:
         logging.info("The database is not yet migrated. Migrating...")
         args = Namespace(
             from_revision=None,
             from_version=None,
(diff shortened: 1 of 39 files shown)


images/airflow/2.10.3/python/mwaa/database/migrate.py
```diff
@@ -40,12 +40,22 @@ def _migrate_db():
         args = Namespace(migration_wait_timeout=1)
         airflow_db_command.check_migrations(args)
         logging.info("The database is already migrated.")
+        args = Namespace(
```
kuyperse, line 43: The addition of these args here isn't present in the 2.10.1 and 2.9.2 changes. Is this intentional?

dhegberg (approved):

images/airflow/2.10.3/python/mwaa/database/migrate.py
michaelfu1029: Yes that should be removed, I added that to test and it thought I deleted it

dhegberg (approved):

kuyperse (approved):

images/airflow/2.10.3/python/mwaa/database/migrate.py
```diff
@@ -45,7 +45,7 @@ def _migrate_db():
         args = Namespace(
             from_revision=None,
             from_version=None,
-            reserialize_dags=None,
+            reserialize_dags=True,
```
FrancisLee13, line 48: Actually, can we not set this flag at all, and just let airflow use whatever value it defaults to?

FrancisLee13: Also, have you confirmed that re-serialization will be triggered with this flag and didn't trigger without the fix?

Issue #129, opened 2024-08-09
Create required log groups if not existing when executing run.sh for local development


ashishgo-aws:
## Overview

When running the Airflow containers via run.sh file, it does not create the log group which each Airflow container reports logs to. This leads to error messages regarding the groups being missing and there are no logs reported to CloudWatch. 

## Acceptance Criteria

The error message is gone and logs are available.

PR #219, opened 2025-02-15
Create log groups in customer account if not exists. #129

sudiptoghosh01:
*Issue #, if available:#129*

*Description of changes:*
Updated the `run.sh` script to check if the log group exists in the customer account or not. If it does not exists, create the required log group for all airflow components.

*Testing*
1. Checked if the log groups were created in the account or not when running the script.

### DIFF
--- a/images/airflow/2.10.1/run.sh
+++ b/images/airflow/2.10.1/run.sh
@@ -48,6 +48,7 @@ export FERNET_KEY
 
 ACCOUNT_ID="" # Put your account ID here.
 ENV_NAME="" # Choose an environment name here.
+REGION="us-west-2" # Keeping the region us-west-2 as default.
 
 # AWS Credentials
 AWS_ACCESS_KEY_ID="" # Put your credentials here.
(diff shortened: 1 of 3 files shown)


images/airflow/2.10.1/python/mwaa/entrypoint.py
```diff
@@ -663,6 +663,44 @@ def run_airflow_command(cmd: str, environ: Dict[str, str]):
         case _:
             raise ValueError(f"Unexpected command: {cmd}")
 
+def _extract_log_group_name(log_group_arn):
+    return log_group_arn.split(":")[-1] if log_group_arn else None
+
+def _create_log_group_if_not_exists(log_group_name):
```
vishalvijay18, line 669: Maybe we should also check if logging is enabled for that airflow component before creating log groups.

sudiptoghosh01: Yeah, that makes sense. I'll make the changes accordingly. Also I noticed it will makes more sense to move these changes to config.py instead of the entry point, so I'll do that as well. Thanks!

Vraj2725 (dismissed):

images/airflow/2.10.1/python/mwaa/logging/config.py
```diff
@@ -177,6 +209,7 @@ def _configure():
     # we run the "dag-processor" Airflow command.
     for comp in ["Worker", "Scheduler", "WebServer", "Triggerer"]:
         log_group_arn, log_level, logging_enabled = get_mwaa_logging_env_vars(comp)
+        _create_log_group_if_not_exists(comp, log_group_arn, logging_enabled)
```
Kytha, line 212: Not sure if this should be the responsibility of the image itself. It may be better to include this logic in the run.sh script that runs prior to launching containers for this repo.

My reasoning for this is the log group is a shared resource across all containers in an MWAA environment. From the perspective of the MWAA service, this logic does not provide value since successful log group creation is a pre-req for launching compute.

Furthermore, this code here will attempt to create log groups for each component, even if the component is just the scheduler. In addition, this code will get executed everytime this config is imported into a python module, which I currently count as 5 times in this codebase. When you consider that any MWAA environment may launch dozens of containers, this will create lots of unnecessary calls to createLogGroup on customer behalf. Throttling on this API in customer account caused my MWAA has been an issue in the past.

Kytha (changes requested):

Vraj2725 (approved):

Kytha (approved):

PR #221, opened 2025-02-21
Graceful updates metrics

ashishgo-aws:
*Description of changes:*

Graceful updates has introduced new logic and edge cases to the image code. This change adds new metrics to be emitted for those edge cases by the task monitor.

### DIFF
--- a/images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
+++ b/images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
@@ -738,19 +738,19 @@ class WorkerTaskMonitor:
         self.stats.incr("mwaa.task_monitor.worker_shutdown_termination_timeout", termination_timeout_metric)
 
         # Report a metric about the number of current task at shutdown, and a warning in case this is greater than zero.
-        # If the worker was marked for killing or was marked for termination and the allowed time limit for termination has been breached,
-        # then these interruptions are expected and another metric is also emitted to signify that.
         interrupted_task_count = self._get_current_task_count()
-        unexpected_interrupted_task_count = 0 \
-            if self.marked_for_kill or self.is_termination_time_limit_breached() else interrupted_task_count
-
         if interrupted_task_count > 0:
             logger.warning("There are non-zero ongoing tasks.")
         self.stats.incr(f"mwaa.task_monitor.interrupted_tasks_at_shutdown", interrupted_task_count)
 
-        if unexpected_interrupted_task_count > 0:
-            logger.warning("Worker was not shutdown via expected methods and some tasks were interrupted.")
-        self.stats.incr(f"mwaa.task_monitor.unexpected_interrupted_tasks_at_shutdown", unexpected_interrupted_task_count)
+        if self.mwaa_signal_handling_enabled:
+            # If the worker was marked for killing or was marked for termination and the allowed time limit for termination
+            # has been breached, then these interruptions are expected and another metric is also emitted to signify that.
+            unexpected_interrupted_task_count = 0 \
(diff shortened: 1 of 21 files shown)


images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
```diff
@@ -290,33 +290,87 @@ def _cleanup_undead_process(process_id: int):
...
-    signal_data["processed"] = True
-    with open(signal_filepath, "w") as file_pointer:
-        json.dump(signal_data, file_pointer)
-    logger.info(f"Successfully processed signal {signal_data['executionId']}")
+    def __str__(self):
+        """
+        String representation of the SignalData object.
+        :return: String containing signal data in JSON format.
+        """
+        return json.dumps({
```
dhegberg, line 331: Can this use self.to_json()?

images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
```diff
@@ -290,33 +290,87 @@ def _cleanup_undead_process(process_id: int):
...
+        """
+        return {
+            'executionId': self.executionId,
+            'signalType': self.signalType.value,
+            'createdAt': self.createdAt,
+            'processed': self.processed
+        }
+
+    @classmethod
+    def from_json_string(cls, json_string: str) -> 'SignalData':
```
dhegberg, line 351: Should this be `from_string` for consistency?

images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
```diff
@@ -474,35 +559,60 @@ def process_next_signal(self):
...
+        match signal_type:
+            case SignalType.ACTIVATION:
+                # Skipping checking if activation is already present because activation signals are by
+                # design sent in redundant fashion to improve success of a newly created worker from getting
+                # activated by the worker enabler lambda.
+                self.waiting_for_activation = False
+            case SignalType.KILL:
+                if self.marked_for_kill:
+                    logger.warning("Received kill signal but already marked for kill. Ignoring.")
+                    signal_ignored = 1
```
dhegberg, line 590: Is the signal ignored in this case? You're re-applying on the line below.

images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
ashishgo-aws: The reason I named it this way was to basically give a hint to the user that the argument needs to be a JSON string. A non json string will result in parsing error. But I don't have a strong preference. Let me know if you want me to change it.

images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
ashishgo-aws: Sure, changed.

images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
ashishgo-aws: My intention behind signal_ignored was to catch the unknown cases where we have redundant signals coming in. Here, we have a kill signal coming in for a worker which has already been marked for killing from a previous kill signal. Hence, the new signal is redundant and we emit signal_ignored = 1.

dhegberg (approved):

images/airflow/2.10.1/python/mwaa/celery/task_monitor.py
dhegberg: I think it's probably fine.

anrajago (approved):

Mercury2699 (approved):

Issue #223, opened 2025-02-21
Use initdb instead of migratedb to handle database initialization/upgrade


FrancisLee13:
## Overview

Currently, the database initialization/upgrade is done by calling airflow's `migratedb` in db_command.py. `migratedb` internally calls airflow's `upgradedb`, which has more parameters than `initdb` does, including some parameters with default values. Using `migratedb` requires us to provide all of parameters, and we also need to make sure the parameters we provide are consistent across airflow versions, whereas using `initdb` we don't need to provide any parameters, and we can depend on the default(which is, in most cases, recommended) behaviour of airflow when doing db migration/initialization.

## Acceptance Criteria

In the next version release, change the `migratedb` call to `initdb` in file `<version>/python/database/migrate.py`

## Additional Info

<Optional if you have additional info/context that you are useful to attach>

FrancisLee13:
`initdb` in `airflow/cli/commands/db_command.py` is deprecated, but the same function in `airflow/utils/db.py` is not marked as deprecated. 

When we work on the next Airflow version release, we should confirm what should be the recommended way.

sudiptoghosh01:
I think `airflow db init` command was deprecated in favor of `airflow db migrate` in airflow 2.7, refer, https://airflow.apache.org/docs/apache-airflow/2.7.2/release_notes.html#the-db-init-db-upgrade-commands-and-database-load-default-connections-configuration-options-are-deprecated-33136

PR #224, opened 2025-03-02
Replace migratedb with initdb for database initialization

golamhisham:
This PR replaces the `migratedb` call with `initdb` in `migrate.py` to ensure the default behavior of Airflow database initialization is used.

This is my first open-source contribution!

### DIFF
--- a/images/airflow/2.10.3/python/mwaa/database/migrate.py
+++ b/images/airflow/2.10.3/python/mwaa/database/migrate.py
@@ -51,7 +51,7 @@ def _migrate_db():
             to_version=None,
             use_migration_files=None,
         )
-        airflow_db_command.migratedb(args)
+        airflow_db_command.initdb(args)
         logging.info("The database is now migrated.")
 
 


MazrimT:
initdb is depricated with the following DepricationWarning:
```
"`db init` is deprecated.  Use `db migrate` instead to migrate the db and/or "
"airflow connections create-default-connections to create the default connections",
```

FrancisLee13:
As mentioned in [issue#223](https://github.com/aws/amazon-mwaa-docker-images/issues/223), we should only change this for next version release, not the existing versions.


> initdb is depricated with the following DepricationWarning

I was thinking of calling `initdb` in `airflow/utils/db.py`, which doesn't have deprecation warning. Although I am not sure if that is also on deprecation path as well and just not marked yet. We should decide when we are working on the next airflow version.

Mercury2699:
Closing due to inactivity, feel free to reopen if there's new discussions.

PR #232, opened 2025-03-21
Raise exception when customer environment variables fail to load.

sudiptoghosh01:
*Issue #, if available:N/A*

*Description of changes:*
Enhanced error handling for environment variable loading failures by adding explicit exception raising instead of silent logging. Previously, errors were only logged which made debugging difficult. This change ensures immediate visibility of configuration issues and helps in faster root cause analysis when environment variables fail to load.

New Behavior:
- If there is a problem with the service code during the loading of customer environment from the startup script, the containers will fail.

*Testing*
Verified container behavior by simulating a failure scenario (`removed JSON import`). Container failed with clear error message as expected.

### DIFF
--- a/images/airflow/2.10.1/python/mwaa/entrypoint.py
+++ b/images/airflow/2.10.1/python/mwaa/entrypoint.py
@@ -340,6 +340,10 @@ def execute_startup_script(cmd: str, environ: Dict[str, str]) -> Dict[str, str]:
                 return customer_env_dict
             except Exception as e:
                 logger.error(f"Error reading the customer's environment variables: {e}")
+                PROCESS_LOGGER.error(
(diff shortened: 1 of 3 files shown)


Vraj2725:
There are other places where exceptions are silently handled, are we planning to cover them also?

sudiptoghosh01:
> There are other places where exceptions are silently handled, are we planning to cover them also?

Will take care of that as part of our second action item. For now, will just raise an exception in case the environment variables fails to load.

Vraj2725 (approved):

vishalvijay18 (dismissed):

Vraj2725 (dismissed):

vishalvijay18 (approved):

Vraj2725 (approved):

PR #234, opened 2025-03-27
Raise exception for missing customer_env_vars.json file.

sudiptoghosh01:
*Issue #, if available:N/A*

*Description of changes:*
Currently, when `customer_env_vars.json` file fails to be created due to service-side issues, the error is only logged silently, which can cause issue later while debugging the main issue. This change makes the container fail explicitly by raising an exception instead of silently continuing execution.

*Testing:*
1. Verified container failure by simulating missing `customer_env_vars.json` file.
2. Confirmed appropriate error message is thrown.


ashishgo-aws (dismissed):

vishalvijay18 (approved):

ashishgo-aws (approved):

Vraj2725 (approved):
LGTM

PR #242, opened 2025-04-08
Adding fix for incorrect log level bug

Vraj2725:
*Issue #, if available:*
When using airflow versions 2.9.2 and above customer are unable to see error logs when they set their loglevel to `WARNING` or above. The root cause is that in subprocess.py, the line `self.process_logger.info(line.decode("utf-8"))` overrides the log level coming from airflow stream to INFO in [subprocess.py](https://github.com/aws/amazon-mwaa-docker-images/blob/main/images/airflow/2.9.2/python/mwaa/subprocess/subprocess.py#L217).

*Description of changes:*
- Setting process logger level to `DEBUG` to emit all logs and setting root logger to customer's log level to respect customer's log level.
- Refactoring `entrypoint.py` to resolve circular dependency issue and breaking down its logic into different files so all files have single responsibility such as:
     - `user_requirements.py`: Installing user requirements.
     - `execute_command.py`: Logic for executing all the subprocesses
     - `setup_environment.py`: Running the startup script and setting up environment variables

*Testing*
- Added unit tests for all refactored code. 
- Ran log heavy dag on the refactored image code by using local docker.
- Exported environment variable using startup script and verified that it set correctly.

This PR is for V 2.9.2 only. We will release the fix for 2.10.1 and 2.10.3 after these changes reach prod.

### DIFF
--- a/images/airflow/2.9.2/python/mwaa/config/setup_environment.py
+++ b/images/airflow/2.9.2/python/mwaa/config/setup_environment.py
@@ -1,0 +1,272 @@
+"""
+Environment configuration module for Amazon MWAA.
+
+This module handles the setup and configuration of environment variables
+and settings required for running Apache Airflow in Amazon MWAA (Managed
+Workflows for Apache Airflow) environments.
+"""
+# Python imports
+from datetime import timedelta
+from typing import Dict
+import json
+import logging
+import os
+import shlex
+import time
+
+# Our imports
+from mwaa.config.airflow import (
+    get_essential_airflow_config,
+    get_opinionated_airflow_config,
+    get_user_airflow_config
+)
+from mwaa.logging.config import MWAA_LOGGERS
+from mwaa.config.environ import get_essential_environ, get_opinionated_environ
+from mwaa.subprocess.subprocess import Subprocess
+from mwaa.subprocess.conditions import TimeoutCondition
+
+
+
+logger = logging.getLogger("mwaa.entrypoint")
+
+STARTUP_SCRIPT_SIGTERM_PATIENCE_INTERVAL = timedelta(seconds=5)
+STARTUP_SCRIPT_MAX_EXECUTION_TIME = (
+    timedelta(minutes=5) - STARTUP_SCRIPT_SIGTERM_PATIENCE_INTERVAL
+)
+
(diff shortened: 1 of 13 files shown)


images/airflow/2.10.1/requirements.txt
```diff
@@ -25,5 +25,7 @@ pyright
 ruff
 sqlalchemy-stubs
 pytest
+pytest-asyncio
```
RishabhBajaj97, line 28: nit: lets keep the scope of this PR limited to 2.9.2, also curious why did we choose to go with 2.9.2 initially?

images/airflow/2.9.2/python/mwaa/config/setup_environment.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+logger = logging.getLogger("mwaa.entrypoint")
+
+STARTUP_SCRIPT_SIGTERM_PATIENCE_INTERVAL = timedelta(seconds=5)
+STARTUP_SCRIPT_MAX_EXECUTION_TIME = (
+    timedelta(minutes=5) - STARTUP_SCRIPT_SIGTERM_PATIENCE_INTERVAL
+)
+
+def _execute_startup_script(cmd: str, environ: Dict[str, str]) -> Dict[str, str]:
```
vishalvijay18, line 37: Sudipto had introduced new exception for startup script failures. Is it getting removed as part of this PR?

Refer: https://github.com/aws/amazon-mwaa-docker-images/commit/7614d8526897fa2bed573cc074e2442199e2f870

Also why it didn't lead to any check failures?

vishalvijay18 (changes requested):

images/airflow/2.9.2/python/mwaa/config/setup_environment.py
Vraj2725: This shouldn't have been removed. I will add this in next commit. Checks passed as I wrote unit tests as per this code's logic. I will update the unit tests accordingly.

RishabhBajaj97:
Can we add the testing details to this PR?

vishalvijay18 (approved):

RishabhBajaj97 (approved):

PR #247, opened 2025-04-24
Add migrate-db command and support airflow downgrades

michaelfu1029:
*Issue #, if available:*

*Description of changes:*
Added a new command `migrate-db` for a special container that will run the same airflow db init/upgrade logic but will also check for an environment variable `MWAA__CORE__TARGET_VERSION` to downgrade. Existing inplace upgrade will not be changed for now.

The migrate_with_downgrade.py file is a copy or migrate.py but with an additional downgrade check

### DIFF
--- a/images/airflow/2.10.1/python/mwaa/database/update.py
+++ b/images/airflow/2.10.1/python/mwaa/database/update.py
@@ -1,0 +1,87 @@
+"""
+This script is responsible for running Airflow meta database migrations. This will replace
+the migrate script.
+
+IMPORTANT NOTE: This script must be run with all the required environments exported,
+just like when running any Airflow command, as it imports Airflow modules and needs to
+connect to the meta database, thus all configurations need to be set.
+"""
+
+from argparse import Namespace
+import logging.config
+import os
+import sys
+
+from mwaa.utils.dblock import with_db_lock
+from airflow.cli.commands import db_command as airflow_db_command
+
+
+# Usually, we pass the `__name__` variable instead as that defaults to the module path,
+# i.e. `mwaa.entrypoint` in this case. However, since this is a script, `__name__` will
+# have the value of `__main__`, hence we hard-code the module path.
+logger = logging.getLogger("mwaa.database.update")
+
+
+def _verify_environ():
+    """
+    This script is supposed to have all the environment variables required for running
+    Airflow, since we will be using Airflow modules directly. This function verifies
+    they are set by ensuring the existence of the `AWS_EXECUTION_ENV`, which we add
+    during the creation of the `environ` dictionary in the entrypoint.py.
+    """
+    if not os.environ.get("AWS_EXECUTION_ENV", "").startswith("Amazon_MWAA_"):
+        logger.error("The necessary environment variables are not set.")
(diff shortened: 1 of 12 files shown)


images/airflow/2.10.1/python/mwaa/database/update.py
```diff
@@ -0,0 +1,87 @@
...
+    they are set by ensuring the existence of the `AWS_EXECUTION_ENV`, which we add
+    during the creation of the `environ` dictionary in the entrypoint.py.
+    """
+    if not os.environ.get("AWS_EXECUTION_ENV", "").startswith("Amazon_MWAA_"):
+        logger.error("The necessary environment variables are not set.")
+        sys.exit(1)
+
+
+@with_db_lock(1234)
+def _migrate_db():
```
kuyperse, line 38: I would rename this file to something more descriptive like migrate_with_downgrade.py rather than update.py.

images/airflow/2.10.1/python/mwaa/database/update.py
```diff
@@ -0,0 +1,87 @@
...
+            to_version=None,
+            use_migration_files=None,
+        )
+        airflow_db_command.migratedb(args)
+        logging.info("The database is now migrated.")
+
+def _check_downgrade_db():
+    target_version = os.environ.get("MWAA__CORE__TARGET_VERSION", None)
+    current_version = os.environ.get("AIRFLOW_VERSION", None)
+    if target_version and current_version and current_version != target_version:
```
kuyperse, line 61: The target_version may be larger in the case of an environment upgrade, therefore this should check whether current_version > target_version. You can use the python module packages. Something like `return version.parse(target_airflow_version) < version.parse(current_version)`

kuyperse (changes requested):

sudiptoghosh01:
Can you also include the corresponding unit test for your code change. I see the quality checks are failing for this change.

sudiptoghosh01 (changes requested):
Can you please include the corresponding unit test for your code changes. I see the quality checks are failing for this change.

michaelfu1029:
Unit tests are added, is only failing because did not add test for 
```
if command == "migrate-db":
        await airflow_db_migrate(environ)
        print("Finished running db validations")
        return
```

but there is already a unit test for `airflow_db_migrate()` so I don't believe this is necessary

kuyperse (approved):

sudiptoghosh01:
> Unit tests are added, is only failing because did not add test for
> 
> ```
> if command == "migrate-db":
>         await airflow_db_migrate(environ)
>         print("Finished running db validations")
>         return
> ```
> 
> but there is already a unit test for `airflow_db_migrate()` so I don't believe this is necessary. I believe we also don't have coverage tests for other commands

I believe you should cover this branch as well. Similar to what we have for other commands, please refer to this [code pointer](https://github.com/aws/amazon-mwaa-docker-images/blob/main/tests/images/airflow/2.9.2/python/mwaa/test_entrypoint_2_9_2.py#L245 ).

kuyperse (dismissed):

sudiptoghosh01 (approved):

Issue #246, opened 2025-04-24
create_venv.py failing due to libpythondylib not found


sdananjay93:
I am using python installed via mise. 
```
➜  git_personal cd amazon-mwaa-docker-images/
➜  amazon-mwaa-docker-images git:(main) ls
CODE_OF_CONDUCT.md   CONTRIBUTING.md      NOTICE               create_venvs.py      pyrightconfig.json   quality-checks       ruff.toml
CODING_GUIDELINES.md LICENSE              README.md            images               pytest.ini           requirements.txt     tests
➜  amazon-mwaa-docker-images git:(main) which python
/Users/dananjay/.local/share/mise/installs/python/3.12.10/bin/python
```

`create_venvs.py` is failing with the following stacktrace
```
➜  amazon-mwaa-docker-images git:(main) python create_venvs.py --target development
>>> Creating a virtual environment under the path .venv...
> Creating virtualenv in directory: .venv
Traceback (most recent call last):
  File "/Users/dananjay/git_personal/amazon-mwaa-docker-images/create_venvs.py", line 151, in <module>
    main()
  File "/Users/dananjay/git_personal/amazon-mwaa-docker-images/create_venvs.py", line 143, in main
    create_venv(
  File "/Users/dananjay/git_personal/amazon-mwaa-docker-images/create_venvs.py", line 52, in create_venv
    venv.create(venv_path, with_pip=True)
  File "/Users/dananjay/.local/share/mise/installs/python/3.12.10/lib/python3.12/venv/__init__.py", line 528, in create
    builder.create(env_dir)
  File "/Users/dananjay/.local/share/mise/installs/python/3.12.10/lib/python3.12/venv/__init__.py", line 77, in create
    self._setup_pip(context)
  File "/Users/dananjay/.local/share/mise/installs/python/3.12.10/lib/python3.12/venv/__init__.py", line 387, in _setup_pip
    self._call_new_python(context, '-m', 'ensurepip', '--upgrade',
  File "/Users/dananjay/.local/share/mise/installs/python/3.12.10/lib/python3.12/venv/__init__.py", line 383, in _call_new_python
    subprocess.check_output(args, **kwargs)
  File "/Users/dananjay/.local/share/mise/installs/python/3.12.10/lib/python3.12/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
[...]

PR #248, opened 2025-04-28
Use symlinks for virtual environment creation to fix ensurepip failure. (#246)

sudiptoghosh01:
*Issue #, if available:#246*

*Description of changes:*
Update virtual environment creation to use symlinks, addressing ensurepip failures.

*Testing*
Validated by running the image and successfully creating a new virtual environment using symlinks.

*Ref*
Known bug for python on MacOs: https://bugs.python.org/issue38705

### DIFF
--- a/create_venvs.py
+++ b/create_venvs.py
@@ -49,7 +49,7 @@ def create_venv(path: Path, development_build: bool, recreate: bool = False):
 
     if not venv_path.exists():
         print(f"> Creating virtualenv in directory: {venv_path}")
-        venv.create(venv_path, with_pip=True)
+        venv.create(venv_path, with_pip=True, symlinks=True)
     else:
         print(f"> Virtualenv already exists in {venv_path}")
 


Vraj2725 (approved):
LGTM.

create_venvs.py
```diff
@@ -49,7 +49,7 @@ def create_venv(path: Path, development_build: bool, recreate: bool = False):
 
     if not venv_path.exists():
         print(f"> Creating virtualenv in directory: {venv_path}")
-        venv.create(venv_path, with_pip=True)
+        venv.create(venv_path, with_pip=True, symlinks=True)
```
dhegberg, line 52: Does this need to be parameterized at the CLI level?

Will `symlinks = True` work in all configurations?

sudiptoghosh01: The `symlinks=True` parameter is safe across all platforms. It's specifically needed to fix a known [issue](https://bugs.python.org/issue38705) with virtual environments on macOS, but does not cause problems on other OS. 

About making this parameterized, yeah we could add a CLI flag, like `--use-symlinks`, if needed, but hardcoding this is a simple solution that works everywhere. Please let me know you thoughts.

dhegberg (approved):

Issue #240, opened 2025-04-03
Move BOM generation to scheduled github action


Kytha:
Updates to the Bill of Materials are causing bloat in PRs. This makes it less clear to reviewers what meaningful changes are being made to source code; it is just annoying. Also, the BOM will only be updated if the requester actually re-builds all the images. 

Instead, it may be a good idea to move generating the BOM behind a docker build argument, like GENERATE_BILL_OF_MATERIALS. Then we can have a daily scheduled GitHub action which runs a build to all images with GENERATE_BILL_OF_MATERIALS = True, and publish a PR with the updated artifacts. 

This would allow us to keep PRs clean while keeping all BOMs up-to-date. This approach would only require approval/merging of the auto-generated PRs, but even this can be automated with auto-merge and review overrides for the bot.

sudiptoghosh01:
PR is merged, closing the issue.

PR #252, opened 2025-05-02
GitHub action to update BOM artifacts at scheduled time interval. (#240)

sudiptoghosh01:
*Issue #, if available: #240*

*Description of changes:*
Adds a scheduled GitHub Actions workflow to automate Bill of Materials (BOM) updates for Airflow images.

*Testing*
Tested by manually triggering the workflow in a forked repo. Confirmed version detection, BOM generation, branch creation, and PR flow all work as expected.


Vraj2725 (dismissed):
Please update Readme on how we can locally update BOM artifacts also.

Vraj2725 (approved):

.github/workflows/update-bom.yaml
```diff
@@ -0,0 +1,129 @@
...
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+          git config --global user.email "action@github.com"
+          git config --global user.name "GitHub Action"
+
+      - name: Create the necessary Python virtual environments
+        run: python3.11 ./create_venvs.py --target production
+
+      - name: Start Docker service
+        run: |
+          dockerd &
```
Kytha, line 55: I'm curious how this is working. My understanding was that since this action is running inside a container, running dockerd would require --privileged, which github actions doesn't support. But your notes say you were able to build image so I assume this isn't an issue?

Kytha (approved):

.github/workflows/update-bom.yaml
sudiptoghosh01: Yeah, I have tested this changes by running these actions in my forked repo, it was working fine.

PR #268, opened 2025-06-05
Updated entrypoint & healthcheck logic  to handle container exists during grace period

samratde:
#### Description of changes :
- Added `try-catch` in `entrypoint.py` which will handle all the `Exception`s thrown, gracefully and create a `container_unhealthy` marker file (flag)
- Updated `healthcheck.py` to check for the above marker file
- Added UTs for `healthcheck.py`
- Added new test for `entrypoint.py`

#### Testing
- Hardcoded an Exception line in the startup script execution block
- Deployed pinwheel environment : The ECS failed to stabilize which in turn trigger Circuit breaker and that resulted in CFN ROLLBACK
    -  CREATE : Iselink for test stack - https://tiny.amazon.com/2l4r798c/IsenLink
    - UPDATE : Isenlink for test stack - https://tiny.amazon.com/jq4npiah/IsenLink

### DIFF
--- a/images/airflow/2.10.1/healthcheck.py
+++ b/images/airflow/2.10.1/healthcheck.py
@@ -4,6 +4,7 @@ Module for running healthcheck on airflow processes.
 
 import subprocess
 import sys
+import os
 from enum import Enum
 
 class ExitStatus(Enum):
@@ -67,6 +68,12 @@ def get_airflow_process_command(airflow_component: str):
     exit_with_status(ExitStatus.INVALID_AIRFLOW_COMPONENT)
 
 if __name__ == '__main__':
+
+    # Check if an 'container_unhealthy' marker file is present
+    if os.path.exists("/tmp/container_unhealthy"):
+        print("/tmp/container_unhealthy file found - marking as unhealthy.")
+        exit_with_status(ExitStatus.AIRFLOW_COMPONENT_UNHEALTHY)
+
     if len(sys.argv) != 2:
         print("Usage: python3 healthcheck.py <airflow_component>")
         exit_with_status(ExitStatus.INSUFFICIENT_ARGUMENTS)
--- a/images/airflow/2.10.1/healthcheck.py
+++ b/images/airflow/2.10.1/healthcheck.py
@@ -69,8 +69,8 @@ def get_airflow_process_command(airflow_component: str):
 
 def main():
     # Check if an 'container_unhealthy' marker file is present
-    if os.path.exists("mwaa/container_unhealthy"):
-        print("mwaa/container_unhealthy file found - marking as unhealthy.")
+    if os.path.exists("/tmp/mwaa/container_unhealthy"):
+        print("/tmp/mwaa/container_unhealthy file found - marking as unhealthy.")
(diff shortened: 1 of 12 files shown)


images/airflow/2.10.1/healthcheck.py
```diff
@@ -67,6 +68,12 @@ def get_airflow_process_command(airflow_component: str):
     exit_with_status(ExitStatus.INVALID_AIRFLOW_COMPONENT)
 
 if __name__ == '__main__':
+
+    # Check if an 'container_unhealthy' marker file is present
+    if os.path.exists("/tmp/container_unhealthy"):
```
uttpal, line 73: Will airflow user also have access to write files in /tmp/?. If yes then user's process may create this file in-intentionally and fail health check .

We should write to a location where customer does not has access to
Or,  make it hard to do this un-intentionally eg. have unique name like aws-mwaa-health-check-failed-0001

images/airflow/2.10.3/python/mwaa/entrypoint.py
```diff
@@ -69,6 +69,29 @@ def _configure_root_logger(command: str):
...
+    # Create a health check marker file
+
+    # max of (service grace period or container health check startPeriod) + health check (retries * interval)
+    # 1020 + 10 * 75 = ~1800
+    max_timeout = 1800
+
+    try:
+        with open("/tmp/container_unhealthy", "w") as f:
+            f.write("Container encountered fatal error\n")
+        # open('/mwaa/container_unhealthy', 'w').close()
```
Mercury2699, line 82: Parity between other versions.

images/airflow/2.10.1/python/mwaa/entrypoint.py
```diff
@@ -69,6 +69,30 @@ def _configure_root_logger(command: str):
...
+    try:
+        os.makedirs('/tmp/mwaa', exist_ok=True)
+        open('/tmp/mwaa/container_unhealthy', 'w').close()
+
+        logger.error("Marked container as unhealthy via /tmp/mwaa/container_unhealthy")
+    except Exception as file_err:
+        logger.error("Failed to write /tmp/mwaa/container_unhealthy: %s", str(file_err))
+
+    container_runtime = time.time() - CONTAINER_START_TIME
+    if container_runtime < max_timeout:
```
michaelfu1029, line 88: Me and Jake are testing the change for the new images and have noticed that if we include the time.sleep() in the check, the circuit breaker doesnt seem to trigger but when I remove that section for time.sleep(), circuit breaker is triggering after around 11 minutes. We are kind of confused of why this is the behaviour and if the sleep is even needed at all.

Kytha: Without sleep, if you are failing your task to soon then circuit breaker is likely still in stage 1 and will trip. Issue is related to stage 2 behaviour - https://docs.aws.amazon.com/AmazonECS/latest/developerguide/deployment-circuit-breaker.html

Kytha (approved):

anrajago (approved):

PR #282, opened 2025-06-24
Fix linting quality checks for Airflow 2.10.3

soumy-jain:
*Issue # (if available):*
N/A

*Description of changes:*
Quality checks currently do not correctly run against files under ./images due to how pyright resolves its configuration from pyrightconfig.json. Specifically, pyright will use the pyrightconfig.json file in your current directory even if you specify a different root path as an argument.

This PR fixes the issue by:
- Updating lint_python.sh to specify the correct pyrightconfig.json path using the -p switch
- Currently testing with Airflow 2.10.3 only, other versions temporarily skipped for validation
- Resolved linting issues in task_monitor.py and webserver_config.py for version 2.10.3
- Changed pyrightconfig.json from strict to standard mode for appropriate linting configuration

This ensures quality checks correctly run against files under ./images and catch invalid code like missing imports before being merged.

*List any breaking changes for*
* *The Environment runtime*: None
* *The local development experience*: None

### DIFF
--- a/images/airflow/2.10.3/pyrightconfig.json
+++ b/images/airflow/2.10.3/pyrightconfig.json
@@ -8,9 +8,8 @@
     // just mark the changes we make, hence we ignore linting it.
     "python/mwaa/celery/sqs_broker.py"
   ],
-  "strict": ["./"],
   "pythonVersion": "3.11",
   "pythonPlatform": "All",
-  "typeCheckingMode": "strict",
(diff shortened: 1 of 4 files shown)


soumy-jain:
@Vraj2725 Added unit tests in commit 7cc69ac. Please check.

Vraj2725 (approved):

PR #313, opened 2025-08-07
Set reserialize_dags to False on DB upgrades and downgrades

michaelfu1029:
*Issue #, if available:*

*Description of changes:*
Set reserialize_dags to False on DB upgrades and downgrades as it is loading dags for the upgrades/downgrades. We do not install requirements for this process now so we cannot load their dags.


kuyperse:
We still need to re-serialize the DAGs in the Scheduler.

vishalvijay18 (dismissed):

kuyperse (approved):

Mercury2699 (approved):

PR #314, opened 2025-08-07
 [PINWHEEL-5408] Logging issue

thadaninilesh:
**_Issue_**
When emitting logs to CloudWatch we validate the log level of logs against the value set by the customer on the console. If the level matches, the log is pushed to CloudWatch. This is done using the below if block:
`if record.levelno >= logging._nameToLevel[os.environ.get( f"MWAA__LOGGING__AIRFLOW_{self.logs_source.upper()}_LOG_LEVEL", "INFO")]: `
This condition works fine for most of the usecases.
However, in case of Scheduler and Worker MWAA overrides the log level coming from Airflow to INFO, which means if Airflow emits ERROR logs, MWAA makes them INFO while reading the logs via a stream and so the logLevel is lost. This is done because the log level is not available for logs coming from Airflow and we read them via a stream.
`self.process_logger.info(line.decode("utf-8"))`

**_Description of changes:_**
To fix this issue, we are now validating the logging level set in the console in the subprocess as well. This is under the verified assumption that only the desired logs will arrive from Airflow (If WARNING is set in console, Airflow will send WARNING and ERROR logs only). We manually set the log level to the LOG_LEVEL set in console.

**This PR is to propagate changes to all remaining versions. PR for 2.10.3 can be found [here](https://github.com/aws/amazon-mwaa-docker-images/pull/304)**


thadaninilesh:
Build is failing due to missing test cases for 3.0.3

Vraj2725 (approved):

anrajago (approved):

anrajago (approved):

Issue #315, opened 2025-08-07
Adding handling of boto response


RishabhBajaj97:
**Describe the bug**
Currently, we are seeing some customer face issues when the celery broker sends messages to the sqs, adding stack trace below:
```
2025-08-01 11:55:00.873,"Traceback (most recent call last):
"
2025-08-01 11:55:00.873,"  File ""/usr/local/airflow/.local/lib/python3.11/site-packages/airflow/providers/celery/executors/celery_executor_utils.py"", line 223, in send_task_to_executor
"
2025-08-01 11:55:00.873,"    result = task_to_run.apply_async(args=[command], queue=queue)
"
2025-08-01 11:55:00.873,"             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
"
2025-08-01 11:55:00.873,"  File ""/usr/local/airflow/.local/lib/python3.11/site-packages/celery/app/task.py"", line 594, in apply_async
"
2025-08-01 11:55:00.873,"    return app.send_task(
"
2025-08-01 11:55:00.874,"           ^^^^^^^^^^^^^^
"
2025-08-01 11:55:00.874,"  File ""/usr/local/airflow/.local/lib/python3.11/site-packages/celery/app/base.py"", line 801, in send_task
"
2025-08-01 11:55:00.874,"    amqp.send_task_message(P, name, message, **options)
"
2025-08-01 11:55:00.874,"  File ""/usr/local/airflow/.local/lib/python3.11/site-packages/celery/app/amqp.py"", line 518, in send_task_message
"
2025-08-01 11:55:00.874,"    ret = producer.publish(
"
2025-08-01 11:55:00.874,"          ^^^^^^^^^^^^^^^^^
"
2025-08-01 11:55:00.874,"  File ""/usr/local/airflow/.local/lib/python3.11/site-packages/kombu/messaging.py"", line 186, in publish
"
2025-08-01 11:55:00.874,"    return _publish(
"
2025-08-01 11:55:00.874,"           ^^^^^^^^^
"
2025-08-01 11:55:00.874,"  File ""/usr/local/airflow/.local/lib/python3.11/site-packages/kombu/connection.py"", line 556, in _ensured
"
2025-08-01 11:55:00.874,"    return fun(*args, **kwargs)
"
2025-08-01 11:55:00.874,"           ^^^^^^^^^^^^^^^^^^^^
"
[...]

PR #319, opened 2025-08-11
Adding logs to debug unexpected response in celery sqs broker

Vraj2725:
*Issue #, if available:*
https://github.com/aws/amazon-mwaa-docker-images/issues/315

*Description of changes:*
Adding a check to make sure `Attributes` field is present and log the response if no to helping debugging the issue

### DIFF
--- a/images/airflow/2.10.3/python/mwaa/celery/sqs_broker.py
+++ b/images/airflow/2.10.3/python/mwaa/celery/sqs_broker.py
@@ -1150,10 +1150,11 @@ class Channel(virtual.Channel):
         resp = c.get_queue_attributes(
             QueueUrl=url, AttributeNames=["ApproximateNumberOfMessages"]
         )
-        if "Attributes" in resp:
+        try:
             return int(resp["Attributes"]["ApproximateNumberOfMessages"])
-        else:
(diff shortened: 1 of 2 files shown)


images/airflow/2.10.3/python/mwaa/celery/sqs_broker.py
```diff
@@ -1150,7 +1150,10 @@ def _size(self, queue):
         resp = c.get_queue_attributes(
             QueueUrl=url, AttributeNames=["ApproximateNumberOfMessages"]
         )
-        return int(resp["Attributes"]["ApproximateNumberOfMessages"])
+        if "Attributes" in resp:
+            return int(resp["Attributes"]["ApproximateNumberOfMessages"])
+        else:
+            logger.error("Unexpected response from SQS get_queue_attributes: %s", resp)
```
uttpal, line 1156: In error cases, this will cause exception/un-intended behaviour in calling function which may cause confusion among existing customers. Instead of if else, lets catch exception, log it and rethrow

RishabhBajaj97 (approved):

uttpal (approved):

PR #338, opened 2025-09-12
Adding logs to debug unexpected response in celery sqs broker for remaining versions and syncing unit tests for all versions

Vraj2725:
Adding logs to debug unexpected response in celery sqs broker for remaining versions and syncing unit tests for all versions

*Issue #, if available:* https://github.com/aws/amazon-mwaa-docker-images/issues/315

*Description of changes:*
Adding a check to make sure Attributes field is present and log the response if no to helping debugging the issue

### DIFF
--- a/images/airflow/2.10.1/python/mwaa/celery/sqs_broker.py
+++ b/images/airflow/2.10.1/python/mwaa/celery/sqs_broker.py
@@ -1158,7 +1158,11 @@ class Channel(virtual.Channel):
         resp = c.get_queue_attributes(
             QueueUrl=url, AttributeNames=["ApproximateNumberOfMessages"]
         )
-        return int(resp["Attributes"]["ApproximateNumberOfMessages"])
+        try:
+            return int(resp["Attributes"]["ApproximateNumberOfMessages"])
+        except Exception:
+            logger.error("Unexpected response from SQS get_queue_attributes: %s", resp)
+            raise
 
     def _purge(self, queue):
         """Delete all current messages in a queue."""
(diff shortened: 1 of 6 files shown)


images/airflow/2.10.1/python/mwaa/celery/sqs_broker.py
```diff
@@ -1158,7 +1158,11 @@ def _size(self, queue):
         resp = c.get_queue_attributes(
             QueueUrl=url, AttributeNames=["ApproximateNumberOfMessages"]
         )
-        return int(resp["Attributes"]["ApproximateNumberOfMessages"])
+        try:
```
anrajago, line 1161: My understanding of the issue is: 

- Airflow (via kombu and MWAA’s custom sqs_broker.py) is making a GetQueueAttributes API call to SQS to check the queue size (ApproximateNumberOfMessages).

- The response payload should always include an Attributes key.

- In this case, the response is missing the Attributes key, which triggers KeyError: 'Attributes' error.

That means the SQS API call succeeded at the HTTP level but returned a response without the Attributes block. That’s abnormal and indicates a transient issue with the SQS API.

This is a known issue in kombu, tracked here: https://github.com/celery/kombu/issues/2142. It has been fixed in kombu 5.5.0, but MWAA images relying on previous versions can run into this issue. The root cause is that kombu’s SQS transport isn’t resilient to missing keys, and MWAA inherits this limitation in the current version.

Therefore, should we add retries here to address this transient issue? If `ApproximateNumberOfMessages` is still not present after n staggered retries, we can throw the exception.

PR #341, opened 2025-09-19
Remove migrate being ran on scheduler and webserver

michaelfu1029:
*Issue #, if available:*

*Description of changes:*
Removing the migrate command from being ran on schedulers and webservers because it is already performed on the migratedb container. This will prevent database corruptions in the future as leftover schedulers/webservers on rollbacks cause this issue.


kuyperse (approved):

anrajago (approved):

kuyperse:
Do we want the scheduler to still perform a dag serialization call on startup now that it is no longer performing one as part of the migration?

Issue #353, opened 2025-10-18
Database not initialized on start - `run.sh` fails to start on image 3.0.6


mtrisic:
**Describe the bug:**

Running run.sh fails to start the Airflow image v3.0.6 for some users.
During startup, the database initialization is skipped, causing missing tables and repeated container crashes.: 

```log
mwaa-306-db  | 
mwaa-306-db  | PostgreSQL Database directory appears to contain a database; Skipping initialization
mwaa-306-db  | 
mwaa-306-db  | 2025-10-18 07:18:50.013 UTC [1] LOG:  starting PostgreSQL 13.22 (Debian 13.22-1.pgdg13+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
mwaa-306-db  | 2025-10-18 07:18:50.015 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
mwaa-306-db  | 2025-10-18 07:18:50.015 UTC [1] LOG:  listening on IPv6 address "::", port 5432
mwaa-306-db  | 2025-10-18 07:18:50.019 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
mwaa-306-db  | 2025-10-18 07:18:50.025 UTC [27] LOG:  database system was shut down at 2025-10-18 07:15:49 UTC
mwaa-306-db  | 2025-10-18 07:18:50.032 UTC [1] LOG:  database system is ready to accept connections
```

After that, containers enter a crash loop with errors such as:

```log
mwaa-306-db         | 2025-10-18 08:15:03.305 UTC [444] ERROR:  relation "log" does not exist at character 13
mwaa-306-db         | 2025-10-18 08:15:03.305 UTC [444] STATEMENT:  INSERT INTO log (dttm, event, owner, extra) VALUES ('2025-10-18T08:15:03.288038+00:00'::timestamptz, 'cli_api_server', 'airflow', '{"host_name": "dbf9b600ee19", "full_command": "[''/usr/local/airflow/.local/bin/airflow'', ''api-server'']"}') RETURNING log.id
mwaa-306-db         | 2025-10-18 08:15:03.336 UTC [445] ERROR:  relation "log" does not exist at character 13
[...]

PR #357, opened 2025-10-23
Add migratedb container to dockerfile for run.sh

michaelfu1029:
*Issue #, if available:*
https://github.com/aws/amazon-mwaa-docker-images/issues/353

*Description of changes:*

Add the migratedb container to the dockerfile as we removed db initialization outside of this container


images/airflow/2.10.1/docker-compose.yaml
```diff
@@ -112,11 +112,24 @@ services:
...
+      postgres:
+        condition: service_healthy
+      sqs:
+        condition: service_healthy
 
   spy:
     <<: *airflow-common
     command: spy
     container_name: mwaa-2101-spy
+    depends_on:
```
kuyperse: I don't think the spy container needs to depend on the migrate-db container.

PR #376, opened 2025-12-16
Psycopg2 operational error fix

yumike2006:
*Description of changes:*
1. Fixed a celery config error from before where pool recycle and pool preping options were not working because of missing top level config: `database_engine_options`
2. Added keep alive args per Airflow [recommendation](https://airflow.apache.org/docs/apache-airflow/3.0.6/howto/set-up-database.html#setting-up-a-postgresql-database) for managed DB use cases. 
3. Also added these configs to RDS health checks as we are seeing false alarms for connection issues due to operational errors.

*Testing:*
Using a stress test dag that frequently reproduced psycopg2 operational errors, the above fix dramatically reduced the error occurrences as well as RDS health check false alarms. However, there are still sporadic `psycopg2.operationalerror: SSL SYSCALL error: EOF detected` errors, but they may be caused by resource constraint since my test environment CPU and memory are at 100% constantly. Removing the keepalive args will cause it to be more frequent.


sudiptoghosh01 (approved):

images/airflow/2.10.1/python/mwaa/config/database.py
```diff
@@ -98,3 +98,13 @@ def get_db_connection_string() -> str:
     creds = f"{postgres_user}:{postgres_password}"
     addr = f"{postgres_host}:{postgres_port}"
     return f"{protocol}://{creds}@{addr}/{postgres_db}?sslmode={postgres_sslmode}"
+
+# Per recommendation from https://airflow.apache.org/docs/apache-airflow/2.10.1/howto/set-up-database.html,
+# since we are using Amazon RDS
+MWAA_CONNECT_ARGS = {
+    "connect_timeout": 15,
```
kuyperse: Any reason in particular we are increasing this value to 15 from 3? Is a higher value recommended for Airflow?

yumike2006: Just to be consistent. For older versions we were using 15. In the context of Airflow and celery higher is better theoretically as it implies more reliability.

kuyperse (approved):

PR #409, opened 2026-02-17
Add --use-uv flag for faster venv creation using uv

budaesandrei:
## Summary

Adds an opt-in `--use-uv` flag to `create_venvs.py` that uses [uv](https://docs.astral.sh/uv/) instead of the stdlib `venv` + `pip` for significantly faster virtual environment creation and package installation.

## Changes

- Added `is_uv_available()` helper to check if `uv` is on PATH.
- Added `uv_install()` function that uses `uv pip install --python <venv_python>` for package installation.
- Updated `create_venv()` to accept a `use_uv` parameter:
  - When `True`: uses `uv venv` for venv creation and `uv pip install` for all packages, skipping the pip upgrade step (not needed with uv).
  - When `False` (default): behavior is completely unchanged.
- Added `--use-uv` argparse flag with a clear error message if uv is not installed.
- Updated README with usage example.

## Compatibility

- CI workflows are unaffected — they call `create_venvs.py --target production` without `--use-uv`.
- `uv venv` produces a standard venv structure (`bin/activate`, `bin/python`) so all downstream consumers (quality checks, linting, unit tests, build scripts, VS Code settings) work identically.
- Requirements files use standard PEP 508 syntax fully supported by uv.
- The flag is purely opt-in — no existing behavior changes without it.

## Testing

Verified the following scenarios:
- `--use-uv` with uv installed — creates venv with `uv venv`, installs packages with `uv pip install`
- Without `--use-uv` — identical behavior to before
- `--use-uv` without uv installed — exits with clear error message
- `--use-uv` with `--recreate` — properly deletes and recreates venv using uv
- `--help` — new flag appears in usage output

### DIFF
--- a/create_venvs.py
+++ b/create_venvs.py
@@ -32,15 +32,24 @@ def verify_python_version():
         sys.exit(1)
 
 
-def create_venv(path: Path, development_build: bool, recreate: bool = False):
+def is_uv_available() -> bool:
+    """Check if uv is available on PATH."""
+    return shutil.which("uv") is not None
+
+
+def create_venv(
+    path: Path, development_build: bool, recreate: bool = False, use_uv: bool = False
+):
     """
     Create a venv in the given directory and optionally recreate it if it already exists.
 
     :param path: The path to create the venv in.
     :param development_build: Is this a development build.
     :param recreate: Whether to recreate the venv if it already exists.
+    :param use_uv: Whether to use uv instead of pip for package installation.
     """
     venv_path = path / ".venv"
+    installer = uv_install if use_uv else pip_install
     print(f">>> Creating a virtual environment under the path {venv_path}...")
 
     if recreate and venv_path.exists():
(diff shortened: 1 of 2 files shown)


create_venvs.py
```diff
@@ -49,28 +58,33 @@ def create_venv(path: Path, development_build: bool, recreate: bool = False):
 
     if not venv_path.exists():
         print(f"> Creating virtualenv in directory: {venv_path}")
-        venv.create(venv_path, with_pip=True, symlinks=True)
+        if use_uv:
+            subprocess.run(["uv", "venv", str(venv_path)], check=True)
+        else:
```
Mercury2699, line 63: nit: Without `--python`, `uv` picks its own default interpreter, which, although unlikely, may not match what `verify_python_version()` validated above — silently creating a `venv` on a different Python version.

Suggested fix:
```python
subprocess.run(["uv", "venv", "--python", sys.executable, str(venv_path)], check=True)
```

create_venvs.py
```diff
@@ -115,6 +132,26 @@ def pip_install(venv_dir: Path, *args: str):
...
+    """
+    subprocess.run(
+        [
+            "uv",
+            "pip",
+            "install",
+            "--python",
+            os.path.join(venv_dir, "bin", "python"),
+            *args,
+        ],
```
Mercury2699, line 150: The existing codebase already has Windows support — the README documents a PowerShell 5.1 setup, and `main` already guards against Windows in `create_venv()`:
```python
# Already in main branch:
venv.create(venv_path, with_pip=True, symlinks=(sys.platform != "win32"))
```
So Windows developers do run `create_venvs.py` locally. This hardcoded path will silently fail for them — Windows venvs use `Scripts\python.exe`, not `bin/python`.

Suggested fix:
```python
python_path = str(Path(venv_dir) / ("Scripts" if sys.platform == "win32" else "bin") / "python")
```

Mercury2699 (commented):
Thank you for your contribution!

Overall this is solid work — the approach is clean and well-scoped. 

The PR also has a merge conflict with main that will need a rebase before this can land.

Happy to approve once the comments and the rebase are addressed.

PR #410, opened 2026-02-17
Replace manual --version validation with argparse choices using dynamic version discovery

budaesandrei:
## Summary

Replaces the manual path-existence validation for the `--version` argument in `create_venvs.py` with argparse's built-in `choices` parameter, populated dynamically from the filesystem.

## Changes

- Added `get_airflow_versions()` function that scans `./images/airflow/` and returns only directory names matching a semver pattern (`^\d+\.\d+\.\d+$`), so non-version folders (e.g. docs, temp directories) are excluded.
- Changed `--version` argument from `type=str` to `choices=available_versions`, delegating validation to argparse.
- Removed the manual `version_path.exists()` check and associated error handling block, since argparse now handles invalid input with a clear error message listing valid choices.

## Testing

Verified the following scenarios:
- `--version 2.10.1` (valid) — works as expected
- `--version` omitted — creates venvs for all versions (unchanged behavior)
- `--version invalid` — argparse rejects with clear error listing available choices

### DIFF
--- a/create_venvs.py
+++ b/create_venvs.py
@@ -32,6 +32,20 @@ def verify_python_version():
         sys.exit(1)
 
 
+def get_airflow_versions() -> list[str]:
+    """Scan ./images/airflow/ and return folder names that look like semantic versions."""
+    airflow_path = Path("./images/airflow")
+    if not airflow_path.is_dir():
+        return []
+    return sorted(
+        [
+            d.name
+            for d in airflow_path.iterdir()
+            if d.is_dir() and re.match(r"^\d+\.\d+\.\d+$", d.name)
+        ]
+    )
+
+
 def create_venv(path: Path, development_build: bool, recreate: bool = False):
     """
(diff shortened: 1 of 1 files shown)


create_venvs.py
```diff
@@ -129,49 +145,36 @@ def main():
...
-        
         project_dirs = [
             Path("."),
-            version_path,
+            Path(f"./images/airflow/{args.version}"),
         ]
     else:
         project_dirs = [
             Path("."),
             *Path("./images").glob("airflow/*"),
```
Mercury2699, line 169: Since we've dynamically discovered the exact valid version directories in `available_versions` above, we could use that list here too instead of the broad glob. This ensures that a full run (without `--version`) processes the exact same subset of directories that are considered valid choices for `--version`.

```python
        project_dirs = [
            Path("."),
            *[Path(f"./images/airflow/{v}") for v in available_versions],
        ]
```

PR #450, opened 2026-04-08
Create Airflow 3.2.0 Image

wangyimosquito:
*Description of changes:*
+ This PR creates the Airflow 3.2.0 image with necessary changes (For clarity each change is separate to standalone commit):
    + **Statsd**: Updated import from `airflow.metrics.statsd_logger` to `airflow._shared.observability.metrics.statsd_logger` because the module was relocated in Airflow 3.2.0.
    + **Celery Config**: Added `AIRFLOW__CELERY__EXTRA_CELERY_CONFIG` with JSON-serialized broker config because Airflow 3.2.0's `create_celery_app()` no longer reads `celery_config_options` and `getsection()` returns all values as strings, breaking nested dict types like `predefined_queues`.
   + **Disable Triggerer queue config**: Disabled triggerer queues (`AIRFLOW__TRIGGERER__QUEUES_ENABLED=False`) because MWAA runs a single triggerer without `--queues` CLI support, and enabling it would cause all deferrable tasks to hang indefinitely.
+ Unit tests are also imported to 3.2.0 with changes accordingly.

*Testing*:
+ Passed all unit tests.
+ Deployed the image to MWAA and tested out basic functionality working.


images/airflow/3.2.0/bin/airflow-user/execute-user-startup-script
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+echo "Executing customer startup script"
```
isshishi, line 2: In the case of a new version image being created during the process of an official release (such as 3.1.6 then 3.2.0, or 2.9.0 -> 2.9.1 -> 2.9.2), instead of copying the latest repo, we move it instead. This way, it becomes easier to verify that the latest worked on changes have also been ported over. 

Its my bad on putting "copy" in the task instead of move. 

Do you think you can rename the 3.1.6 folder instead of copying it? This way the files don't lose their 3.1.6 specific change history and it becomes easier to verify the changes. 

For example: https://github.com/aws/amazon-mwaa-docker-images/pull/70/changes/c8f7b06c1846808a53c6dd233c5995005f75f82a#diff-e2eae87ac540b9824bd56f73cbda781ed46adf5794001f378408dedff306e411

wangyimosquito: Got it, I thought we are going to keep the 3.1.6 image here. Can redo with `git mv`.

images/airflow/3.2.0/python/mwaa/config/airflow.py
```diff
@@ -212,6 +212,18 @@ def _get_essential_aiflow_execution_api_config() -> Dict[str, str]:
         "AIRFLOW__EXECUTION_API__JWT_EXPIRATION_TIME": "86400"
```
isshishi, line 197: nit: just like this function has a basic test like [this](https://github.com/wangyimosquito/amazon-mwaa-docker-images/blob/37eb0ae35aea48ef2c5d078dedd7a35ccce38080/tests/images/airflow/3.2.0/python/mwaa/config/test_airflow.py#L235), the addition below should have a basic config setting test too.

wangyimosquito: Updated this commit with dedicated config test.

images/airflow/3.2.0/python/mwaa/config/celery.py
```diff
@@ -36,6 +36,7 @@ def create_celery_config() -> dict[str, Any]:
                                   "default": {"url": get_sqs_queue_url()}},
             "is_secure": should_use_ssl(),
             "region": get_aws_region(),
+            "visibility_timeout": 43200,
```
isshishi: The PR description says this Airflow 3.2.0's create_celery_app() no longer reads celery_config_options. So I was thinking why even bother updating this, and instead just remove celery_config_options. However, the module-level `app` ([here](https://github.com/apache/airflow/blob/v3-2-stable/providers/celery/src/airflow/providers/celery/executors/celery_executor_utils.py#L80)) still uses `get_celery_configuration()` which reads `celery_config_options`. Workers execute tasks under this `app`, so `celery.py` is still active on the worker side.

Since both paths need consistent broker config, we should either consolidate to one place or add comments in both to prevent drift.

images/airflow/3.2.0/python/mwaa/config/airflow.py
```diff
@@ -37,14 +38,40 @@ def _get_essential_airflow_executor_config(executor_type: str) -> Dict[str, str]
...
+            # from getsection("celery_broker_transport_options") which returns all
+            # values as strings, breaking predefined_queues (which must be a dict).
+            # It also ignores celery_config_options entirely.
+            #
+            # To work around this, we use extra_celery_config which is parsed via
+            # getjson() and properly preserves nested dict types. This config is
+            # merged into the Celery config dict built by get_default_celery_config().
+            from mwaa.celery.sqs_broker import Transport
+            from mwaa.utils import qualified_name
+            logger.info("********* MWAA extra_celery_config fix applied for Airflow 3.2+ *********")
```
isshishi: The asterisks will be too glaring too see in the logs. Probably a good idea to remove them.

isshishi (approved):

harsh-bhalani (approved):

PR #452, opened 2026-04-10
Add Windows PowerShell 5.1 support for local runner

mailtobash:
Add support for powershell scripts, so these images can be run on windows machines. 

*Issue #, if available:*

*Description of changes:*


images/airflow/3.1.6/run.sh
```diff
@@ -55,6 +55,21 @@ export REGION
 AWS_ACCESS_KEY_ID="" # Put your credentials here.
 AWS_SECRET_ACCESS_KEY="" # Put your credentials here.
 AWS_SESSION_TOKEN="" # Put your credentials here.
+
+# If no real credentials are provided, use dummy values and disable CloudWatch logging
+if [ -z "$AWS_ACCESS_KEY_ID" ]; then
```
isshishi: nit: 

Taking a look at `create_log_group_if_not_exists` [here](https://github.com/aws/amazon-mwaa-docker-images/blob/5f5fb67ce43658898da5d930079f239a355682c6/images/airflow/3.1.6/run.sh#L123), log group is created if $ENV_NAME above has been set. 

So the practical application of this is if someone set the `$ENV_NAME` parameter above but forgot to set the creds here, correct? If so, a visible error from the aws logs call seems more helpful than silently disabling all logging. This way the user would at least know they forgot credentials. Silently turning off logging could lead to confusion later when they expect to see logs in CloudWatch and nothing is there.

The run.ps1 approach handles this better IMO. It sets dummy credentials but leaves logging enabled. ElasticMQ and local services don't validate creds, so everything starts fine. CloudWatch call fail gracefully container-side. And if someone actually configures `$ENV_NAME` expecting CloudWatch, the log group creation still fails visibly at the aws logs call, which is the right signal. Logging config stays truthful.                                                                                                                                                                                   
                                                                                                                                                                                                                  Whichever approach we go with, it should be applied consistently across all versions.

images/airflow/3.2.0/build.ps1
isshishi, line 1: We are about to migrate 3.1.6 to 3.2.0. If that change comes out first, please migrate this to 3.2.0 instead.

images/airflow/3.1.6/run.sh
mailtobash: Done - the silent-disable block has been removed from all versions. All run.sh files now use dummy "local" credentials by default and keep logging enabled, consistent with the run.ps1 approach. The 3.2.0 version is also covered following the upstream rename from 3.1.6.

isshishi (approved):

Mercury2699 (approved):

PR #460, opened 2026-04-13
Enable reading of triggerer execution logs in Task Log UI for 3.x

SergioAAI:
*Description of changes:*
Add triggerer logs in the Task log UI with it's matching task

When logs are loading in the UI, we now query for triggerer execution logs that match the task displayed. If found , triggerer logs are sent back with the task logs , and both are displayed in the UI

# Testing

Unit tests added

Example of triggerer execution logs in UI ( the sleeping logs ) in AF 3.2:

image

### DIFF
--- a/images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
+++ b/images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
@@ -438,7 +438,7 @@ class CloudWatchRemoteTaskLogger(BaseLogHandler, LoggingMixin):
         all_logs: LogMessages = []
         for stream_name in triggerer_streams:
             all_logs.append(
-                f"*** Reading triggerer logs from: {stream_name}"
+                f"Reading triggerer logs from: {stream_name}"
             )
             try:
                 events = self.get_cloudwatch_logs(stream_name, task_instance)
@@ -446,7 +446,7 @@ class CloudWatchRemoteTaskLogger(BaseLogHandler, LoggingMixin):
                     all_logs.append(StructuredLogMessage.model_validate(log))
             except Exception as e:
                 all_logs.append(
-                    f"*** Failed to read triggerer logs from: {stream_name}: {e}"
+                    f"Failed to read triggerer logs from: {stream_name}: {e}"
                 )
         return all_logs
 
@@ -466,7 +466,7 @@ class CloudWatchRemoteTaskLogger(BaseLogHandler, LoggingMixin):
             triggerer_logs = self._read_triggerer_logs(triggerer_streams, task_instance)
         except Exception as e:
             triggerer_logs = [
-                f"*** Failed to read triggerer logs: {e}"
+                f"Failed to read triggerer logs: {e}"
             ]
 
         return messages + logs + triggerer_logs, metadata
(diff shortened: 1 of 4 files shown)


images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
```diff
@@ -395,7 +460,16 @@ def read(
...
         stream_name = self._render_filename(task_instance, try_number).replace(":", "_")
         messages, logs = self._read_remote_logs(stream_name, task_instance)
-        return messages + logs, metadata
+
+        try:
+            triggerer_streams = self._discover_triggerer_streams(stream_name)
+            triggerer_logs = self._read_triggerer_logs(triggerer_streams, task_instance)
+        except Exception as e:
+            triggerer_logs = [
+                f"*** Failed to read triggerer logs: {e}"
```
wangyimosquito: Nit: might be better to remove the "*", it makes it looks more like a left over test-related logging. Same for the above "*** Failed to read triggerer logs from".

wangyimosquito (approved):

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
SergioAAI: Removed the prefix ***, so that the logging is uniform

kuyperse (approved):

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
```diff
@@ -395,7 +460,16 @@ def read(
         """
         stream_name = self._render_filename(task_instance, try_number).replace(":", "_")
         messages, logs = self._read_remote_logs(stream_name, task_instance)
-        return messages + logs, metadata
+
+        try:
+            triggerer_streams = self._discover_triggerer_streams(stream_name)
+            triggerer_logs = self._read_triggerer_logs(triggerer_streams, task_instance)
```
ashish-kumar-goel, line 466: Discover triggerer streams is designed to scan all streams based on prefix and then read triggerer logs execute log query on each of those streams. Both these combined can lead to multiple cloudwatch log query executing and cause throttling for the customer.

Is the prefix here ensuring that only a single or very low number of streams will be returned? If not, how can we ensure that? If yes, can we put guardrails in these methods so that they cannot be misused.

SergioAAI: Yes the prefer returns a single stream , `{base_task_stream_name}.trigger.`, with the full log stream being `{base_task_stream_name}.trigger.N.log` with N being the triggerer job id,  so this function would only return either 1 or 0 log streams when the customer is in the UI. I don't think this should lead to throttling for the customer.

ashish-kumar-goel: Makes sense. Just a small nitpick, you can make this change whenever time permits, but it will then be better if "discover trigger streams" is renamed to suggest that it will return at most a single stream and then you also have checks in the method about the length of the output list (OR maybe it returns a single string instead of list OR maybe the method can have an input parameter which is the max length of the output list and for this use case, the caller of the method can provide 1 as the maxSize)

SergioAAI: Noted, this can be added for clarity in a follow-up PR

PR #493, opened 2026-05-07
Fix filtering log levels for Airflow 3.2.1+ structlog

SergioAAI:
# Description
Airflow 3.1+ uses structlog which outputs all levels to stdout, ignoring `AIRFLOW_CONSOLE_LOG_LEVEL`. The previous code only controlled which logger method to call but never filtered lines below the threshold.

This adds `_parse_log_level()` to extract severity from both structlog (`[info     ]`) and legacy (`{module.py:123} INFO -`) formats, then drops lines below the configured threshold in `_read_subprocess_log_stream()`.

# Testing
Unit tests rewritten to cover parsing and filtering behavior.
Verified with local image.

Example of filtering for warning: 
image

### DIFF
--- a/images/airflow/3.2.1/python/mwaa/subprocess/subprocess.py
+++ b/images/airflow/3.2.1/python/mwaa/subprocess/subprocess.py
@@ -61,12 +61,14 @@ _LEVEL_MAP = {
 }
 
 
-def _parse_log_level(line: str) -> int:
+def _parse_log_level(line: str) -> Optional[int]:
     """
     Parse the log level from a subprocess output line.
 
-    Supports both structlog format and standard Airflow log format.
-    Returns logging.INFO if the level cannot be determined.
+    Supports both structlog and standard Airflow log formats. Returns ``None``
+    when no level prefix is present so the caller can inherit the last
+    recognized level (for example, a traceback body that follows an ``[error]``
+    header).
     """
     # Try structlog format first (most common in Airflow 3.1+)
     m = _STRUCTLOG_LEVEL_RE.search(line)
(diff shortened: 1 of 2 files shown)


wangyimosquito (dismissed):

isshishi (commented):
Just want to verify if we have verified behavior change between versions. If we haven't, can you check what happens when you run a DAG with mixed log levels (log.debug(..), log.info, log.warning, log.error, and a raised exception) with the scheduler log level set to WARNING? Specifically:

1. Do ERROR lines get logged as ERROR and WARNING lines as WARNING? (Previous versions would label everything uniformly as WARNING.)
2. Are traceback lines following an error preserved, or do they get dropped because they lack a level prefix?

Comparing against 3.0.6 with the same log level config would be ideal to confirm the behavior matches.

SergioAAI:
> Just want to verify if we have verified behavior change between versions. If we haven't, can you check what happens when you run a DAG with mixed log levels (log.debug(..), log.info, log.warning, log.error, and a raised exception) with the scheduler log level set to WARNING? Specifically:
> 
>     1. Do ERROR lines get logged as ERROR and WARNING lines as WARNING? (Previous versions would label everything uniformly as WARNING.)
> 
>     2. Are traceback lines following an error preserved, or do they get dropped because they lack a level prefix?
> 
> 
> Comparing against 3.0.6 with the same log level config would be ideal to confirm the behavior matches.

While Task related logs were already working compared to the other logs which had the struct log issue, this did bring the issue the multi-level log line should not default to INFO. This now will take the latest recognized log level, and added tests to reflect this

wangyimosquito (approved):

isshishi (approved):

PR #495, opened 2026-05-08
Fix Non-Critical Logging deadlock issue caused by usage of non fork-safe locks

chgao28:
*Description of changes:*

Fix Non-Critical Logging deadlock issue in Airflow that's caused by the usage of non fork-safe locks that comes with FluentHandler. The issue is more observable when CloudWatch is disabled - and logs are unable to be sent to CloudWatch - this increases the likelihood that the process is holding onto a lock when logs are being re-emitted.

The interaction with celery creating children processes for running tasks and the Fluentbit Sender grabbing non-safe locks caused tasks to be stuck in RUNNING. This is because fork copies the lock state, but does not copy any other threads, causing the child process to not have any thread to unblock.

This PR also adds the functionality and unit tests of Non-Critical Logging to 2.9.2 and 2.10.1 as well, as it was previously missing.

### DIFF
--- a/images/airflow/2.11.0/python/mwaa/logging/cloudwatch_handlers.py
+++ b/images/airflow/2.11.0/python/mwaa/logging/cloudwatch_handlers.py
@@ -12,10 +12,7 @@ import logging
 import os
 import re
 import sys
-import threading
 import traceback
-import weakref
-from queue import Queue
 
 # 3rd party imports
 from airflow.models.taskinstance import TaskInstance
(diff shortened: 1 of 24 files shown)


images/airflow/2.11.0/python/mwaa/logging/cloudwatch_handlers.py
```diff
@@ -12,7 +12,10 @@
 import os
 import re
 import sys
+import threading
```
kuyperse: Unused imports: threading, weakref, and queue. These should be removed.

kuyperse (changes requested):

SergioAAI (approved):

kuyperse (approved):

isshishi (approved):

PR #500, opened 2026-05-20
Add oversized log chunking for task logs in Airflow 3.0.6

harsh-bhalani:
*Issue # (if available):*
https://app.asana.com/1/8442528107068/project/1213077669682433/task/1214436608875623

## Description

Add support for splitting oversized task log messages (>260KB) into
multiple CloudWatch log events, preserving all content. Previously,
oversized logs were silently truncated by watchtower with data loss.

Implementation uses character-level binary search to find optimal chunk
boundaries that respect CloudWatch's per-event size limit (262KB).
Size measurement uses ensure_ascii=True to match watchtower's
CloudWatchLogFormatter serialization, ensuring non-ASCII characters
(CJK, accented text, emojis) are correctly accounted for at 6 bytes
each (\uXXXX escaping).

Design decisions documented inline:
- Character-level binary search chosen (Chosen, optimal fill, O(N·log(N/K)))
- Byte-level splitting rejected (fragile escape-boundary handling)
- Fixed char budget rejected (wasteful for ASCII-heavy content)

Observability:
- Worker log line on each split with original size and chunk count
- StatsD metric: mwaa.logging.task.oversize_split

Testing:
- Unit tests including test verifying chunks fit within
  watchtower's limit with heavy non-ASCII content
- Integration tested on MWAA 3.0.6 environment with mixed content
  oversized logs (512KB-1.2MB), confirmed complete data preservation
  across chunk boundaries with no truncation warnings

## Files Changed

```
images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
tests/images/airflow/3.0.6/python/mwaa/logging/test_cloudwatch_handler.py
```

*List any breaking changes for*
* *The Environment runtime*: N/A
* *The local development experience*: N/A

## Tests
Please refer all the DAGs and their respective logs here - [Test Evidence](https://drive.corp.amazon.com/personal/hbhalani/MWAA/Ops/Oversized_Cloudwatch_Logs_AF_3)

### DIFF
--- a/images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
+++ b/images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
@@ -281,6 +281,7 @@ class CloudWatchRemoteTaskLogger(BaseLogHandler, LoggingMixin):
         task-sdk/src/airflow/sdk/log.py#L143
     """
     LOG_SOURCE = "task"
+    _MAX_EVENT_BYTES = 260_000 # ~254 KB, leaving headroom for CW overhead + JSON envelope
     # In Airflow currently there are some logs that "seeps" out of the traditional task log stream. Ideally we should
     # use log_filename_template config to generate a pattern for filtering only task log streams. But that config
     # value is not a regex pattern but instead a Jinja template. As a workaround for now we use a deny list instead.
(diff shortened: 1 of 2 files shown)


images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
```diff
@@ -317,6 +318,156 @@ def _init_handler(self):
             create_log_group=False,
         )
 
+    def _spilt_oversize_event(self, msg: dict) -> list[dict]:
```
kuyperse: typo: _spilt_ should be _split_

kuyperse (changes requested):

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
harsh-bhalani: oops. will change it.

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
```diff
@@ -317,6 +318,156 @@ def _init_handler(self):
...
+                if isinstance(event_value, str) and len(event_value.encode("utf-8")) > self._MAX_EVENT_BYTES:
+                    metadata = {k: v for k, v in msg.items() if k != "event"}
+                    # Rough truncation — leave room for metadata + truncation marker
+                    truncated = event_value[:self._MAX_EVENT_BYTES // 4] + "\n... [TRUNCATED due to split error] ..."
+                    return [{**metadata, "event": truncated}]
+            except Exception:
+                pass
+            return [msg]
+
+    def _spilt_oversize_event_inner(self, msg: dict) -> list[dict]:
```
chgao28: just reminder that typo is here too.

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
```diff
@@ -317,6 +318,156 @@ def _init_handler(self):
...
+                # [1:-1] strips the quotes to get just the escaped content
+                escaped_size = len(json.dumps(candidate)[1:-1].encode("utf-8"))
+                if escaped_size <= max_event_json_bytes:
+                    best = mid
+                    lo = mid + 1
+                else:
+                    hi = mid - 1
+
+            chunk_str = event_value[pos:pos + best]
+            chunks.append({**metadata, "event": chunk_str})
```
chgao28, line 459: Consider adding a `chunk_index` as part of metadata -> This should be for ordering purposes - since there's an off chance that logs are out of order when uploaded to cloudwatch, and these information can help the customer order their logs. Also consider adding num_chunks to metadata to let the customer know the number of chunks in the logs itself, and not at the end from the print statement.

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
harsh-bhalani: Thanks, updated in new commit.

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
harsh-bhalani: I considered this but decided against it:

1/ CloudWatch preserves event order within a log stream : chunks share the same timestamp and are emitted sequentially, so ordering is guaranteed by insertion order
2/ Adding chunk_index/num_chunks to the JSON introduces fields into the customer's log schema that they didn't emit, which could break downstream parsing tools (SIEM, log aggregators)
3/ I feel chunking should be transparent to the customer

It implicitly suggests customers need to reassemble chunks, but adding extra metadata that is an internal implementation detail they shouldn't need to interact with.

Thoughts?

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
harsh-bhalani: changed

kuyperse (approved):

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
chgao28: I completely agree with the possibility of breaking downstream tools if introducing additional fields. I just digged a bit deeper and you are right that the order are based on the emit order. Therefore, this is more like a "nice-to-have", that would make ordering more obvious. 
Instead, a NIT would be to perhaps add the "[chunk_num/num_chunks]" as part of the logging itself?

chgao28 (approved):

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
harsh-bhalani: Yes definitely it's nice to have, but I'd let customer raise this request, if they want, we can add that. It shouldn't be a big lift.

PR #513, opened 2026-06-11
Fix 3.2.1 to 3.0.6 downgrade FAB table missing default sequence

wangyimosquito:
*Issue #, if available:*
+ **Symptom**: After downgrading from 3.2.1 to 3.0.6 and add new DAG files, new DAGs won't show up in console, and dag processing logs show
```
ERROR - Creation of Permission View Error: (psycopg2.errors.NotNullViolation) null value in column "id" of relation "ab_permission_view" violates not-null constraint

DETAIL: Failing row contains (null, 1, null).

[SQL: INSERT INTO ab_permission_view (permission_id, view_menu_id) VALUES (%(permission_id)s, %(view_menu_id)s) RETURNING ab_permission_view.id]

[parameters: {'permission_id': 1, 'view_menu_id': None}]
``` 
*Description of changes:*
+ Fix FAB (Flask-AppBuilder) security tables after downgrading from Airflow 3.2.1
    (providers-fab 3.6.1 / SQLAlchemy 2.0) to Airflow 3.0.6 (providers-fab 2.4.1 /
    SQLAlchemy 1.4). After the downgrade, the ab_* tables may be left
    without a DEFAULT on the 'id' column. This causes NOT NULL violations because
    SQLAlchemy 1.4's ORM omits the 'id' from INSERT statements, relying on the database
    to auto-generate it via a sequence default.
+ The fix is to add default sequence if missing in FAB tables. The list of tables got from: https://github.com/dpgaspar/Flask-AppBuilder/blob/v4.6.3/flask_appbuilder/security/sqla/models.py
+ Fix also run the `airflow fab-db downgrade` command to handle other mismatches.

*Testing:*
+ [rev1] Deploy the image and created an environment and perform a downgrade, migrate log shows adding sequence to all listed tables, and new DAGs are able to be parsed and loaded in console.
Screenshot 2026-06-11 at 12 36 09 PM

+ [rev2] Deploy the image and created an environment and performed a downgrade, migrate log shows fab-db downgrade successful, and shows default sequence adding to all listed tables. New DAGs are able to be parsed and loaded in console.

### DIFF
--- a/images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py
+++ b/images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py
@@ -89,6 +89,130 @@ def _migrate_db():
         airflow_db_command.migratedb(args)
         logging.info("The database is now migrated.")
 
+def _downgrade_fab_db():
+    """
+    Downgrade the FAB (Flask-AppBuilder) provider database migrations.
+
+    When downgrading from Airflow 3.2.1 (providers-fab 3.6.1) to Airflow 3.0.6
+    (providers-fab 2.4.1), the FAB provider's own migrations need to be reverted.
+    This is equivalent to running: airflow fab-db downgrade --to-version 1.4.0
+
+    The 3.5.0 migration (0001_3_5_0_fix_fab_db_inconsistencies.py) added NOT NULL
+    constraints on ab_permission_view.permission_id and view_menu_id, among other
+    changes. Reverting this ensures compatibility with providers-fab 2.4.1.
+    """
+    from airflow.providers.fab.auth_manager.cli_commands import db_command as fab_db_command
+
+    try:
+        logging.info("Downgrading FAB provider database to version 1.4.0...")
+        fab_args = Namespace(
+            to_version="1.4.0",
+            to_revision=None,
+            from_revision=None,
+            from_version=None,
+            show_sql_only=None,
+            yes=True,
+        )
+        fab_db_command.downgrade(fab_args)
+        logging.info("FAB provider database downgrade completed successfully.")
+    except Exception as e:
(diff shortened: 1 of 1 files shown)


kuyperse:
Have we considered using the airflow fab-db downgrade command instead of modifying the tables ourselves? https://airflow.apache.org/docs/apache-airflow-providers-fab/stable/cli-ref.html#downgrade. We may also need to integrate the migrate command into the upgrade logic as well for Airflow 3.x versions.

wangyimosquito:
> Have we considered using the airflow fab-db downgrade command instead of modifying the tables ourselves? https://airflow.apache.org/docs/apache-airflow-providers-fab/stable/cli-ref.html#downgrade. We may also need to integrate the migrate command into the upgrade logic as well for Airflow 3.x versions.

I've tested the downgrade with pure `airflow fab-db downgrade` method like below:
```
 try:
        logging.info("Downgrading FAB provider database to version 1.4.0...")
        fab_args = Namespace(
            to_version="1.4.0",
            to_revision=None,
            from_revision=None,
            from_version=None,
            show_sql_only=None,
            yes=True,
        )
        fab_db_command.downgrade(fab_args)
        logging.info("FAB provider database downgrade completed successfully.")
```
And this command does not resolve the missing default sequence in id field in  `ab_permission_view` and `ab_view_menu`. [downgrade source code: https://github.com/apache/airflow/blob/main/providers/fab/src/airflow/providers/fab/migrations/versions/0001_3_5_0_fix_fab_db_inconsistencies.py#L321]

kuyperse (approved):

wangyimosquito:
> We may also need to integrate the migrate command into the upgrade logic as well for Airflow 3.x versions.

A fair point, I'll investigate and add a task to add the upgrade command.

SergioAAI (approved):

PR #530, opened 2026-06-25
Fix 3.2.1 deferred-task serialization crash on 3.0.6/2.11.0 downgrade

wangyimosquito:
*Issue #, if available:*
task_instance.next_kwargs (raw JSONB) and trigger.kwargs (airflow.sdk.serde) are written by 3.2.1 but read with BaseSerialization (ExtendedJSON) by the 3.0.6/2.11.0 scheduler and triggerer, which expect a {__var,__type} envelope and have no fallback -> KeyError <Encoding.VAR:'__var'> in get_running_dag_runs_to_examine (scheduler) and Trigger.bulk_fetch (triggerer).

*Description of changes:*
Re-encode both columns into BaseSerialization format BEFORE the Alembic schema downgrade, while the 3.2.1 schema is still present (task_instance.id exists; the Airflow 2.x table has a composite PK with no id column). Data-preserving: in-flight deferred tasks survive the downgrade instead of crashing.

*Testing:*
Deploy the updated image as 3.2.1 environment, triggered a long deferred task before downgrade. Downgrade to 3.0.6 and 2.11.0 successfully. Deferred task is able to maintain the deferred state after downgrade.

### DIFF
--- a/images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py
+++ b/images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py
@@ -40,19 +40,20 @@ def _verify_environ():
         logger.error("The necessary environment variables are not set.")
         sys.exit(1)
 
+@with_db_retry
+def _create_engine_with_retry():
+    engine = create_engine(
+        get_db_connection_string(),
+        **MAINTENANCE_ENGINE_KWARGS,
+    )
+    with engine.connect() as conn:
+        conn.execute(text("SELECT 1"))
+    return engine
+
+
 def _ensure_rds_iam_user():
     try:
-        @with_db_retry
-        def _connect_static():
-            engine = create_engine(
-                get_db_connection_string(),
-                **MAINTENANCE_ENGINE_KWARGS,
-            )
-            with engine.connect() as conn:
-                conn.execute(text("SELECT 1"))
-            return engine
-
-        db_engine = _connect_static()
+        db_engine = _create_engine_with_retry()
         with db_engine.connect() as conn:
             with conn.begin():
                 result = conn.execute(text("SELECT 1 FROM pg_roles WHERE rolname = :rolename"), {"rolename": DB_IAM_USERNAME})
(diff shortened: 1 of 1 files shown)


images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py
```diff
@@ -222,10 +222,125 @@ def _fix_fab_sequence_defaults():
...
+    except Exception as e:
+        logger.error(f"Error while re-encoding deferred-task serialization: {e}")
+        raise
+
+
 def _check_downgrade_db():
     target_version = os.environ.get("MWAA__DB__AIRFLOW_TARGET_VERSION", None)
     current_version = os.environ.get("AIRFLOW_VERSION", None)
     if target_version and current_version and Version(target_version) < Version(current_version):
+        is_legacy_target = Version(target_version) in (Version("3.0.6"), Version("2.11.0"))
```
klisitsynaws: There will be 2.11.2 version. I guess this _fix_deferred_task_serialization() can be done unconditionally.

I believe "legacy" doesn't have special meaning here.

images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py
```diff
@@ -222,10 +222,125 @@ def _fix_fab_sequence_defaults():
...
+        if serde_deserialize is not None:
+            try:
+                obj = serde_deserialize(raw)  # serde-encoded -> python obj
+            except Exception:
+                obj = raw  # already a plain value (e.g. defer-time raw dict)
+        return BaseSerialization.serialize(obj)
+
+    try:
+        @with_db_retry
+        def _connect():
```
klisitsynaws: May be better call this _get_engine()?

klisitsynaws (commented):
left two minor comments

images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py
wangyimosquito: Here it is called `_connect_static()` - https://github.com/aws/amazon-mwaa-docker-images/blame/main/images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py#L46C13-L46C28

But if this is the case I should probably reuse function.

images/airflow/3.2.1/python/mwaa/database/migrate_with_downgrade.py
wangyimosquito: 1. I haven't tested the downgrade path to 2.11.2 yet, so I didn't include 2.11.2 here. It will be better to only include 3.0.6 and 2.11.0 in this PR to resolve issue first, and then add 2.11.2 once full testing is done. (Asana task already created)
2. If `legacy` is confusing I can update the var name in rev2.

klisitsynaws (approved):

isshishi (approved):

PR #531, opened 2026-06-26
Fix log-level filtering dropping non-Airflow subprocess output for 3.2.1

SergioAAI:
# Description

Airflow 3.2.1 introduced structlog-aware log-level filtering in `_read_subprocess_log_stream` (PR #493). This filtering correctly drops lines below the configured `AIRFLOW_CONSOLE_LOG_LEVEL` threshold for Airflow components (scheduler, worker, triggerer) which emit structlog-formatted output.

However, non-Airflow subprocesses — pip install, startup script, and post-startup verification — emit plain text that never matches the structlog (`[info ]`) or legacy Airflow (`} INFO -`) format regexes. Since `_parse_log_level()` returns `None` for these lines, they inherit `last_level=INFO` and get silently dropped when the threshold is WARNING or higher. This means customers lose visibility into requirements installation output and startup script execution when they raise their console log level.

**Fix:** Add an `always_publish` flag (default `False`) to the `Subprocess` class. When `True`, all output lines are published unconditionally via `.info()`, bypassing log-level filtering entirely. The flag is set on the three utility subprocess callers:

- `user_requirements.py` — pip install
- `setup_environment.py` — startup script execution
- `setup_environment.py` — post-startup verification

Airflow component subprocesses (scheduler, worker, triggerer, etc.) retain the default `always_publish=False` and continue using the structlog-aware filtering as intended.

# Testing

- 24 unit tests pass (23 existing + 1 new)
- New test: `test_always_publish_bypasses_log_level_filter` — verifies that with `AIRFLOW_CONSOLE_LOG_LEVEL=ERROR` and `always_publish=True`, plain-text subprocess output is still logged via `.info()`
- Verified locally against Airflow 3.2.1 with `AIRFLOW_CONSOLE_LOG_LEVEL=WARNING`:

| Subprocess | With fix | Without fix |
|---|---|---|
| `requirements_install` | All pip output visible | pip INFO lines silently dropped |
| `startup_script_execution` | All startup messages visible | Only caller timing message remains |
| `post-startup verification` | Verification output visible | Output dropped |

### DIFF
--- a/images/airflow/3.2.1/python/mwaa/subprocess/subprocess.py
+++ b/images/airflow/3.2.1/python/mwaa/subprocess/subprocess.py
@@ -97,7 +97,8 @@ class Subprocess:
         conditions: List[ProcessCondition] = [],
         sigterm_patience_interval: timedelta = _SIGTERM_DEFAULT_PATIENCE_INTERVAL,
         on_sigterm: Optional[Callable[[], None]] = None,
-        is_essential: bool = False
+        is_essential: bool = False,
+        always_publish: bool = False
     ):
         """
         Initialize the Subprocess object.
@@ -133,6 +134,7 @@ class Subprocess:
         self.conditions = conditions
         self.sigterm_patience_interval = sigterm_patience_interval
         self.on_sigterm = on_sigterm
+        self.always_publish = always_publish
 
         self.start_time: float | None = None
         self.process: Popen[Any] | None = None
@@ -267,6 +269,9 @@ class Subprocess:
                     time.sleep(_SUBPROCESS_LOG_POLL_IDLE_SLEEP_INTERVAL.total_seconds())
             else:
                 decoded_line = line.decode("utf-8", errors="replace").rstrip()
+                if self.always_publish:
+                    self.process_logger.info(decoded_line)
+                    continue
                 parsed_level = _parse_log_level(decoded_line)
                 if parsed_level is not None:
                     last_level = parsed_level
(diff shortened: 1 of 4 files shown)


images/airflow/3.2.1/python/mwaa/subprocess/subprocess.py
```diff
@@ -267,6 +267,9 @@ def _read_subprocess_log_stream(self, process: Popen[Any]):
                     time.sleep(_SUBPROCESS_LOG_POLL_IDLE_SLEEP_INTERVAL.total_seconds())
             else:
                 decoded_line = line.decode("utf-8", errors="replace").rstrip()
+                if not self.is_essential:
```
isshishi: The code works, but it couples two unrelated concepts, which could lead to silent bugs.

`is_essential` is not really related to log formatting, and is instead acting as a middle-man proxy. It basically means "restart the ECS task if this subprocess dies", which is unrelated to the log format. It only works correctly today because the essential subprocesses (scheduler/worker/webserver/triggerer) happen to be the structlog emitters and the non-essential ones (pip, startup) happen to be plain text. Nothing enforces that, so it regresses silently the moment someone adds a non-essential structlog subprocess (leaks DEBUG/INFO) or an essential plain-text one (over-filtered). If that happens, debugging also becomes difficult due to naming. 

Prefer an intent-named flag, defaulting to `False`.

```python
def __init__(self, *, cmd, ..., always_publish: bool = False):
    self.always_publish = always_publish

### def _read_subprocess_log_stream:
     ...
     if self.always_publish or line_level >= configured_level:
         ... emit ... 
```

Using something like always_publish removes the proxy by naming the actual decision and setting it explicitly per call site (just the pip and startup invocations), so a new subprocess defaults to the safe filtered path instead of inheriting a bypass by accident.

Also remember to add `always_publish = True` to the pip subprocess and the startup script subprocess.

SergioAAI: Makes sense to have a specific flag for this then. 
I did start with a named flag instead, but then saw that the is_essential flag matched the sub processes we were wanting to let through. But it is better to only use it for it's intended purpose

Added the new flag in the newest change

isshishi (approved):

wangyimosquito (approved):

PR #557, opened 2026-07-14
Change Task logs to use Fluentbit for Airflow 3

chgao28:
## Summary

  Route Airflow 3 task logs through FluentBit when Non-Critical Logging (NCL) is enabled, completing the NCL migration for all log types in 3.0.6 and 3.2.1.

  ## Problem

  When `USE_NON_CRITICAL_LOGGING=true`, all subprocess-based logs (scheduler, worker, webserver, triggerer) and DAG processing logs are already routed through FluentBit via `BaseLogHandler.create_cloudwatch_handler()`. However, **task logs** — which in Airflow 3 use a separate code path via the `CloudWatchRemoteTaskLogger.processors` property and Structlog — were hardcoded to
  watchtower via `get_handler()`.

  Task logs were the only log type still making direct CloudWatch API calls with NCL enabled, bypassing FluentBit's buffering and the non-critical failure semantics. In Airflow 2.x, task logs already route through FluentBit under NCL (`TaskLogHandler.set_context`); this PR brings Airflow 3 to parity.

  ## Solution

  Branch the `processors` property on `self.NON_CRITICAL_LOGGING_ENABLED`:

  - **NCL enabled**: Create a `ForkSafeFluentHandler` with tag `customer.task.logs` (matching the 2.x task handler), attach a `_TaskRoutingFormatter` that serializes the structlog event dict as JSON and includes `log_group`/`log_stream` routing fields for FluentBit delivery.
  - **NCL disabled**: Preserve the existing watchtower path unchanged (including the 3.2.1 `shutting_down` reset for Airflow 3.1+ triggerer compatibility).

  ### Key design decisions

  1. **Non-blocking under back-pressure**: The handler uses `queue_maxsize=50000, queue_circular=True` — identical to the 2.x task handler and the `BaseLogHandler` NCL path. When FluentBit stalls and the queue fills, the oldest event is dropped instead of blocking the task's log call. Combined with the fork-safe sender, logging can never hang or deadlock an Airflow process.

  2. **Handler lifecycle**: The fluent handler is assigned to `self.handler` so Airflow's end-of-task `upload()` → `flush()` and `close()` drain the sender queue before process exit. The sender's delivery thread is a daemon thread, so without an explicit `close()` the final log events of a task (completion status, fatal errors) could be lost.

  3. **Non-critical error handling**: The FluentBit path swallows emit exceptions (increments `mwaa.logging.task.emit_error` only). The watchtower path continues to re-raise. Logging failures must not fail
  tasks.

  4. **JSON serialization**: `json.dumps(record.msg, default=str)` when the message is a dict, mirroring watchtower's `CloudwatchLogFormatter`, so structured logs render correctly in the CloudWatch console. Falls back to `getMessage()` for plain strings.

  5. **Timestamp preservation**: The structlog `timestamp` field is parsed into `record.created`, which the fluent handler uses as the event time — CloudWatch shows the original event time, not delivery time (consistent with watchtower behavior).

  6. **Oversize splitting (3.0.6)**: The fluent path calls `_split_oversize_event()` to chunk >260KB events, preserving the behavior added in #500. The splitter's `ensure_ascii=True` size accounting matches the formatter's `json.dumps` default.

  ## Known gaps (out of scope, follow-ups)

  - **3.2.1 oversize handling**: 3.2.1 has no `_split_oversize_event` (the #500 chunking was only added to 3.0.6). Oversized events in 3.2.1 are passed through and rejected downstream by CloudWatch — pre-existing gap shared with the watchtower path, to be addressed by porting #500 to 3.2.1 separately.
  - **Delivery-failure observability**: The fluent sender reports failures via return value (not exceptions) and its circular-queue drops are silent, so the `emit_error` metric undercounts. This is pre-existing behavior shared with the 2.x NCL path and watchtower's async delivery; a follow-up will check emit return values and wire a `queue_overflow_handler` drop metric across all versions.

  ## Testing

  - **Unit tests** (3.0.6 and 3.2.1):
    - FluentBit handler used when NCL=true, with exact constructor args (`customer.task.logs`, `queue_maxsize=50000`, `queue_circular=True`); watchtower not called
    - Watchtower used when NCL=false; FluentBit not called
    - `queue_circular` propagation through `ForkSafeFluentHandler` to the underlying sender
  - **Manual testing** (dev environment, NCL=true):
    - Task logs delivered to the CloudWatch `-Task` log group with correct JSON formatting and accurate event timestamps
    - Deferrable/triggerer DAG task logs route correctly
    - `dag_processor/` streams filtered from the Task log group (IGNORED_PATTERNS)
    - Stress test: 50 concurrent tasks × 5000 log lines (250k total) — PENDING

  ## Files Changed

  | File | Change |
  |------|--------|
  | `images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py` | NCL branching in `processors`; fluent handler with circular queue, lifecycle wiring, JSON formatting, oversize splitting |
  | `images/airflow/3.2.1/python/mwaa/logging/cloudwatch_handlers.py` | Same, minus oversize splitting (see Known gaps); preserves `shutting_down` reset in watchtower fallback |
  | `tests/images/airflow/{3.0.6,3.2.1}/.../test_cloudwatch_handler.py` | Replace "always watchtower" test with per-path tests for NCL true/false |
  | `tests/images/airflow/{3.0.6,3.2.1}/.../test_fork_safe_fluent_handler.py` | Add `queue_circular` propagation test |

### DIFF
--- a/images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
+++ b/images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
@@ -479,51 +479,113 @@ class CloudWatchRemoteTaskLogger(BaseLogHandler, LoggingMixin):
             handlers. And only the processor attribute from the remote logging class is loaded into the Structlog
             logger used for task logging.
         """
-        from logging import getLogRecordFactory
-
         import structlog.stdlib
-
-        logRecordFactory = getLogRecordFactory()
-        # The handler MUST be initted here, before the processor is actually used to log anything.
-        # Otherwise, logging that occurs during the creation of the handler can create infinite loops.
-        _handler = self.get_handler()
         from airflow.sdk.log import relative_path_from_logger
 
-        def proc(logger: structlog.typing.WrappedLogger, method_name: str, event: structlog.typing.EventDict):
-            if not logger or not (stream_name := relative_path_from_logger(logger)):
+        if self.NON_CRITICAL_LOGGING_ENABLED:
+            _handler = ForkSafeFluentHandler(
+                'customer.task.logs',
+                host='localhost',
+                port=24224,
+                queue_maxsize=50000,
+                queue_circular=True,
+            )
+            log_group = self.log_group_name
+            # Stash log_stream on the handler so the formatter can access it per-call
+            _handler._current_log_stream = ""
+
+            class _TaskRoutingFormatter(logging.Formatter):
+                def format(self, record) -> Dict[str, str]:  # type: ignore[override]
+                    return {
+                        'log_group': log_group,
+                        'log_stream': _handler._current_log_stream,
+                        'message': json.dumps(record.msg, default=str) if isinstance(record.msg, dict) else record.getMessage(),
+                    }
+
+            _handler.setFormatter(_TaskRoutingFormatter())
+            # Assign to self.handler so close() and upload() -> flush() drain the
+            # sender queue at task exit. The sender's send thread is a daemon thread,
+            # so without a proper close() the last log events of a task can be lost.
+            self.handler = _handler
+
+            def proc(logger: structlog.typing.WrappedLogger, method_name: str, event: structlog.typing.EventDict):
+                if not logger or not (stream_name := relative_path_from_logger(logger)):
+                    return event
+                _handler._current_log_stream = stream_name.as_posix().replace(":", "_")
+
+                if self.log_group_name.endswith("-Task") \
+                        and any(re.match(p, _handler._current_log_stream) for p in CloudWatchRemoteTaskLogger.IGNORED_PATTERNS):
+                    return event
+
+                name = event.get("logger_name") or event.get("logger", "")
+                level = structlog.stdlib.NAME_TO_LEVEL.get(method_name.lower(), logging.INFO)
+                if level < self.log_level:
+                    return event
+
+                msg = copy.copy(event)
+                created = None
+                if ts := msg.pop("timestamp", None):
+                    with contextlib.suppress(Exception):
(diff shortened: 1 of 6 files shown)


SergioAAI (approved):

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
```diff
@@ -479,52 +479,114 @@ def processors(self) -> tuple[structlog.typing.Processor, ...]:
...
+                        'message': json.dumps(record.msg, default=str) if isinstance(record.msg, dict) else record.getMessage(),
+                    }
+
+            _handler.setFormatter(_TaskRoutingFormatter())
+            # Assign to self.handler so close() and upload() -> flush() drain the
+            # sender queue at task exit. The sender's send thread is a daemon thread,
+            # so without a proper close() the last log events of a task can be lost.
+            self.handler = _handler
+
+            def proc(logger: structlog.typing.WrappedLogger, method_name: str, event: structlog.typing.EventDict):
```
michaelfu1029, line 511: Nit: Looks mostly the same as the same function below. Would it be possible to combine it into one function that handles the different cases?

michaelfu1029 (approved):

images/airflow/3.0.6/python/mwaa/logging/cloudwatch_handlers.py
chgao28: This is left intentional since the watchtower path and the fluentbit path are different - this allow the two delivery paths to be independently modifiable, minimizing risk on the legacy WatchTower path

PR #574, opened 2026-07-30
Add RDS IAM credential rotation for Airflow 3.x

Mercury2699:
## Summary

Ports RDS IAM credential rotation from the 2.x images to Airflow 3.0.6, 3.2.1, and 3.3.0. This enables IAM-based authentication for metadata DB connections when connecting through RDS Proxy.

The 2.x implementation (`mwaa/utils/get_rds_iam_credentials.py`, `mwaa/config/airflow_rds_iam_patch.py`) is restructured into the 3.x `mwaa.config` package.

**Scope:** this is a port, so it stays behaviourally equivalent to 2.x. Several defects that exist identically in the 2.x images were found while reviewing it; those are listed under Known gaps and are being fixed in cross-version follow-ups rather than only for 3.x. The only intentional divergences are the ones 3.x structurally requires, called out below.

### New modules (per version)

- `mwaa/config/rds_iam_credentials.py` — `RDSIAMCredentialProvider` with thread-safe token caching (5-minute refresh buffer on the 15-minute token expiry), ECS task-credential fetching, boto3 RDS token generation, and connection URL construction from the `MWAA__DB__POSTGRES_*` environment variables.
- `mwaa/config/rds_iam_patch.py` — SQLAlchemy `Engine` `do_connect` event listener that intercepts metadata DB connections and swaps in a fresh IAM token.

### How the patch is loaded

`airflow_local_settings.py`, installed to `${AIRFLOW_USER_HOME}/config` by the Dockerfile, installs the patch — the same mechanism as 2.x. Airflow appends `$AIRFLOW_HOME/config` to `sys.path` and `import_local_settings()` imports `airflow_local_settings` from `settings.initialize()`, which `airflow/__init__.py` runs on every `import airflow`. That makes it the only correct install site: the patch registers a SQLAlchemy `do_connect` listener, and a listener only affects the process that registered it. The Airflow components are spawned by the entrypoint as separate `subprocess.Popen(["airflow", ...])` children with a fresh interpreter, so installing from the entrypoint alone leaves every process that actually talks to the metadata database unpatched.

As in 2.x, customer-provided `dags/airflow_local_settings.py` and `plugins/airflow_local_settings.py` are loaded so their side effects still apply. See Known gaps for the limitation this carries over.

### Intentional divergences from 2.x

Only where 3.x requires it:

- **`make_url` import.** 3.0.6 ships SQLAlchemy 1.4, where `make_url` is only importable from `sqlalchemy.engine`; 3.2.1 and 3.3.0 ship SQLAlchemy 2.0 and import it from `sqlalchemy`.
- **`cparams.clear()`.** SQLAlchemy 2.0's psycopg2 dialect puts `dbname` in `cparams` while `translate_connect_args()` returns `database`. Both present makes psycopg2 raise `you can't specify both 'database' and 'dbname' arguments`, so the listener clears `cparams` before rebuilding from the IAM URL. Not needed on 1.4.
- **`install_rds_iam_patch()` as a function** rather than 2.x's import side effect, which is what makes it testable. It carries an idempotency guard so repeated calls cannot stack listeners, matching the natural idempotency of 2.x's module import.
- **Lazy, fail-safe metadata URL.** 2.x resolves the metadata URL at module import time and raises if the `MWAA__DB__*` variables are absent, which would abort any Airflow process. Here it is resolved on demand, cached, and returns `None` rather than raising into the `do_connect` listener.
- **Explicit failure when no credential path works.** `_ensure_rds_iam_user()` tries static credentials then falls back to an IAM token; if neither opens a connection it raises a `RuntimeError` chained to the underlying error, failing migrate-db and therefore the container. Without a connection the grants cannot be applied, so continuing would bring the environment up in an unknown permission state. 2.x logs and skips. This is deliberate — the unguarded fallback was introduced by this port, so it had to be addressed here either way, and failing loudly was chosen over restoring the silent skip.

### Activation

Off by default. The patch installs only when all of the following hold:

- `USE_IAM_CREDENTIALS=true`
- `MWAA__DB__POSTGRES_SSLMODE=require` (set only for RDS Proxy environments; IAM auth is unsupported without the proxy)
- the process is not `migrate-db`, which uses admin credentials for schema migrations
- the process is not the local runner (`MWAA_LOCAL_RUNNER`)

## Revision history

The later commits fix the initial port, which was not functional. Reviewers looking at the first commits in isolation should note:

- The original port installed the patch from `entrypoint.py`, in the wrong process. An earlier version of this description presented that as an intentional improvement over the 2.x import side effect; that was incorrect, and the `airflow_local_settings.py` override it replaced was load-bearing.
- `_get_metadata_url()` read `AIRFLOW__DATABASE__SQL_ALCHEMY_CONN`. That variable is only placed in the environment dict handed to the Airflow subprocesses and is absent from `os.environ`, so it returned `None`, `_is_accessing_metadata_db()` returned `False`, and token injection was skipped even where the listener was present. It now derives from `get_db_connection_string()`, as 2.x does.
- The `is_local_runner()` guard, the IAM branch in `dblock.py`, and `GRANT adminuser TO airflow_user` had all been dropped.
- The IAM fallback in `_ensure_rds_iam_user()` was unguarded, so a failure propagated out of a function whose every other path logs and returns. A reply on an earlier revision defended this as 2.x parity; that was inaccurate — 2.x wraps the whole body in one `try/except`.

The branch also contains add/revert pairs where fixes for cross-version defects were implemented and then reverted once we settled on keeping this PR a straight port. Those are the Known gaps below.

## Testing

- 36 RDS IAM tests per version (108 total) pass on all three versions. Full suites: 351 on 3.0.6, 370 on 3.2.1, 369 on 3.3.0.
- Verified on SQLAlchemy 1.4.54 (3.0.6) and 2.0.51 (3.2.1, 3.3.0).
- `ruff check .` clean as CI runs it, and clean against each image's own config.

Coverage includes ECS credential retrieval, token generation and caching/refresh, connection URL construction, metadata DB detection via both `cargs` and `cparams`, every `install_rds_iam_patch()` decision branch, and the credential-failure semantics in `_ensure_rds_iam_user()`.

Regression tests cover each defect listed above and were confirmed to fail against the pre-fix code, individually rather than in aggregate. The pre-existing tests mocked `_get_metadata_url` wholesale, which is why the original defect shipped green.

## Known gaps — cross-version, tracked for follow-up

All of these exist identically in the five 2.x images. They are deliberately not fixed here so this PR stays a port and the fixes land uniformly.

- **Customer cluster policies are not registered.** Because `$AIRFLOW_HOME/config` precedes the plugins folder on `sys.path`, this file wins the `import airflow_local_settings` that Airflow performs, and Airflow reads policy names off the module it imported. The customer's file is executed but its names are never re-exported, so `task_policy`, `dag_policy`, `pod_mutation_hook` and settings overrides such as `STATE_COLORS` are silently dropped; only side effects apply. Verified against Airflow 3.0.6: with no MWAA file present the customer's policy registers with 1 hookimpl, with it present the count is 0. This is a regression introduced when `config/airflow_local_settings.py` was added in #381/#392, and #429 was intended to prevent it but only restored execution, not registration.
- **The ECS task credential fetch is unbounded and unretried.** `urllib` is called with no `timeout` and `socket.getdefaulttimeout()` is `None`, so it can block indefinitely — inside the `do_connect` listener. There is no retry, and `with_db_retry` does not help since it only covers `OperationalError` and `InterfaceError`. Per the Fargate Data Plane analysis in P481207036, connect timeouts to `169.254.170.2` are caused by client-side resource contention (a TMDS throttle breach would return HTTP 429 on a completed connection), so the mitigation is client-side timeouts and retries. Note botocore's `AWS_METADATA_SERVICE_TIMEOUT` does not apply, as this fetch is plain `urllib`.
- **Token refresh sits on the connection critical path.** The 15-minute lifetime with a 10-minute refresh exists so a refresh can happen off the critical path, but `get_token()` performs the fetch on the calling thread and holds the lock across it, so a slow fetch blocks every caller in the process. The refresh should be handed off while the still-valid cached token is served immediately.
- **`create_db_connection_url()` bypasses the token cache** on its default `token=None` path. No caller uses that path today, so it is a latent trap rather than an active regression.

Two further items, not cross-version:

- **`cparams.clear()` and non-URL connect args.** `AIRFLOW__DATABASE__SQL_ALCHEMY_CONNECT_ARGS` points at `mwaa.config.database.MWAA_CONNECT_ARGS` (`connect_timeout` and the `keepalives*` settings), which arrive via `cparams` rather than the URL. Clearing `cparams` drops them on IAM-patched connections. Worth confirming whether they need re-applying.
- **Grants are skipped when only IAM works.** If static credentials fail and the IAM fallback succeeds, the connection is as `airflow_user`, so the `current_user` gate skips every grant and the function returns successfully. Pre-existing 2.x behaviour.

For the 2.x follow-up, note that 2.9.2 relies on a `PYTHONPATH` override forcing `config` ahead of `dags`, because Airflow 2.9's single `prepare_syspath()` adds the DAGs folder first; removing it there would stop the patch loading at all. No equivalent is needed in 3.x, where no Airflow version puts the DAGs folder on `sys.path`.

## Compatibility

Backward compatible. With the feature flag unset there is no behavior change: `install_rds_iam_patch()` returns early and no SQLAlchemy listener is registered.

Note that `MWAA_LOCAL_RUNNER` is not forwarded into containers by `docker-compose`, in either 2.x or 3.x, so that guard is inert inside containers. The effective protection for local runs is `MWAA__DB__POSTGRES_SSLMODE: "prefer"` in `docker-compose.yaml`, which makes `is_using_rds_proxy()` false.

### DIFF
--- a/images/airflow/3.0.6/airflow_local_settings.py
+++ b/images/airflow/3.0.6/airflow_local_settings.py
@@ -1,0 +1,62 @@
+"""Airflow local settings configuration for MWAA RDS IAM authentication.
+
+Airflow imports this module during ``airflow.settings.initialize()``, which runs
+unconditionally on ``import airflow``. ``airflow.settings`` appends
+``$AIRFLOW_HOME/config`` to ``sys.path`` before doing so, which is where the
+Dockerfile installs this file.
+
+This is the only correct place to install the RDS IAM credential rotation patch:
+the patch registers a SQLAlchemy ``do_connect`` listener, and that listener lives
+in the process that registered it. The Airflow components (scheduler, worker,
+api-server, dag-processor, triggerer and task runners) are spawned by the
+entrypoint as separate ``subprocess.Popen(["airflow", ...])`` children, each a
+fresh Python interpreter. Installing the patch from the entrypoint alone would
+therefore leave every process that actually talks to the metadata database
+unpatched.
+"""
+
+import importlib.util
+import logging
+import os
+
+# Airflow copies every public name of this module into the `airflow.settings`
+# namespace. Declaring an empty `__all__` keeps our internals from clobbering
+# Airflow's own globals; we export no policies or settings overrides.
+__all__: list[str] = []
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from mwaa.config.rds_iam_patch import install_rds_iam_patch
+
+    install_rds_iam_patch()
+except Exception as e:
+    _logger.error(f"Failed to load RDS IAM patch: {e}")
+    raise
+
+# Load ${AIRFLOW_HOME}/dags/airflow_local_settings.py and
+# ${AIRFLOW_HOME}/plugins/airflow_local_settings.py if they exist, so that
+# customer-provided local settings continue to be honoured.
+_AIRFLOW_HOME = os.environ.get("AIRFLOW_HOME", "/usr/local/airflow")
+_CUSTOMER_LOCAL_SETTINGS = [
+    (
+        "customer_dags_airflow_local_settings",
+        os.path.join(_AIRFLOW_HOME, "dags", "airflow_local_settings.py"),
+    ),
+    (
+        "customer_plugins_airflow_local_settings",
+        os.path.join(_AIRFLOW_HOME, "plugins", "airflow_local_settings.py"),
+    ),
+]
+
+for _module_name, _path in _CUSTOMER_LOCAL_SETTINGS:
+    if not os.path.exists(_path):
+        continue
+    try:
+        _spec = importlib.util.spec_from_file_location(_module_name, _path)
+        if _spec and _spec.loader:
+            _module = importlib.util.module_from_spec(_spec)
+            _spec.loader.exec_module(_module)
+    except Exception as e:
+        _logger.error(f"Failed to load airflow_local_settings from {_path}: {e}")
+        raise
--- a/images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py
+++ b/images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py
@@ -18,6 +18,10 @@ import logging
 import os
 from functools import cache
 
+# NOTE: SQLAlchemy 1.4 (which Airflow 3.0.6 ships) does not re-export
+# make_url from the top-level sqlalchemy package; that alias was only added
+# in SQLAlchemy 2.0. The 3.2.1/3.3.0 images use the preferred
+# `from sqlalchemy import make_url`.
 from sqlalchemy.engine import make_url
 
 from mwaa.config.database import get_db_connection_string
@@ -141,11 +145,17 @@ def install_rds_iam_patch() -> None:
         token = RDSIAMCredentialProvider.get_token()
         url = make_url(RDSIAMCredentialProvider.create_db_connection_url(token))
 
-        # Clear existing connection params to avoid conflicts between
-        # SQLAlchemy versions (SA 2.0 uses 'dbname' in cparams while
-        # translate_connect_args returns 'database' -- both present causes
-        # psycopg2 to raise "can't specify both 'database' and 'dbname'").
-        cparams.clear()
+        # Drop only the credential/identity keys before rebuilding from the
+        # IAM URL. SQLAlchemy 2.0's psycopg2 dialect places 'dbname' in
+        # cparams while url.translate_connect_args() returns 'database', and
+        # psycopg2 rejects a connect() call carrying both (same for
+        # 'username' vs 'user'), so both aliases of each pair are popped.
+        # Everything else in cparams is preserved: engine connect_args such
+        # as MWAA_CONNECT_ARGS' connect_timeout and keepalives* settings
+        # (wired in via AIRFLOW__DATABASE__SQL_ALCHEMY_CONNECT_ARGS) arrive
+        # through cparams and must survive the token injection.
+        for stale_key in ("dbname", "database", "username", "user", "password"):
+            cparams.pop(stale_key, None)
         cparams.update(url.translate_connect_args(username="user"))
         cparams.update(url.query)
 
--- a/images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py
+++ b/images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py
@@ -1,0 +1,111 @@
+"""RDS IAM authentication patch for SQLAlchemy connections.
+
+Attaches a global listener to SQLAlchemy's Engine class that intercepts
+metadata database connections and replaces credentials with fresh IAM tokens.
+
+This module is feature-flag controlled by USE_IAM_CREDENTIALS=true,
+only activates for RDS Proxy environments, and skips the migrate-db process.
+"""
+
+import logging
+import os
+
+from sqlalchemy import make_url
+
+from mwaa.config.rds_iam_credentials import (
+    RDSIAMCredentialProvider,
+    is_using_rds_proxy,
+    use_iam_credentials,
+)
+
+logger = logging.getLogger(__name__)
+
+MWAA_AIRFLOW_COMPONENT = os.environ.get("MWAA_AIRFLOW_COMPONENT")
+
+
+def _is_from_migrate_db() -> bool:
+    """Check if the current process is the migrate-db container."""
+    if MWAA_AIRFLOW_COMPONENT is None:
+        logger.debug("MWAA_AIRFLOW_COMPONENT not set in environment.")
+        return False
+    return MWAA_AIRFLOW_COMPONENT == "migrate-db"
+
+
+def _get_metadata_url():
+    """Get the parsed metadata DB URL from environment."""
+    conn_str = os.environ.get("AIRFLOW__DATABASE__SQL_ALCHEMY_CONN")
+    if not conn_str:
+        return None
+    return make_url(conn_str)
+
+
+def _is_accessing_metadata_db(dialect, cargs, cparams) -> bool:
+    """Return True if the connection targets the Airflow metadata DB."""
+    metadata_url = _get_metadata_url()
+    if metadata_url is None:
+        return False
+
+    # Try detecting from cargs (dsn string)
+    if cargs:
+        try:
+            url_obj = make_url(cargs[0])
+            if (
+                url_obj.host == metadata_url.host
+                and url_obj.database == metadata_url.database
+                and url_obj.port == metadata_url.port
+                and url_obj.username in ["airflow_user", "adminuser"]
+            ):
+                return True
+        except Exception:
+            pass
+
+    # Try detecting from cparams (kwargs dict)
+    host = cparams.get("host")
+    db = cparams.get("dbname") or cparams.get("database")
+    port = cparams.get("port")
+    username = cparams.get("username") or cparams.get("user")
+
+    return (
+        host == metadata_url.host
+        and db == metadata_url.database
+        and port == metadata_url.port
+        and username in ["airflow_user", "adminuser"]
+    )
+
+
+def install_rds_iam_patch() -> None:
+    """Install the RDS IAM authentication patch if conditions are met.
+
+    Conditions:
+    - USE_IAM_CREDENTIALS=true
+    - Environment uses RDS Proxy (MWAA__DB__POSTGRES_SSLMODE=require)
+    - Current process is NOT migrate-db
+    """
+    if not use_iam_credentials():
+        logger.info("RDS IAM patch not installed: USE_IAM_CREDENTIALS is not true.")
+        return
+
+    if not is_using_rds_proxy():
+        logger.info("RDS IAM patch not installed: not using RDS Proxy.")
+        return
+
+    if _is_from_migrate_db():
+        logger.info("RDS IAM patch not installed: running in migrate-db process.")
+        return
+
+    from sqlalchemy import event
+    from sqlalchemy.engine import Engine
+
+    def patch_rds_iam_authentication(dialect, conn_rec, cargs, cparams):
+        """SQLAlchemy do_connect event listener that injects IAM tokens."""
+        if not _is_accessing_metadata_db(dialect, cargs, cparams):
+            return
+
+        token = RDSIAMCredentialProvider.get_token()
+        url = make_url(RDSIAMCredentialProvider.create_db_connection_url(token))
+
+        cparams.update(url.translate_connect_args(username="user"))
+        cparams.update(url.query)
+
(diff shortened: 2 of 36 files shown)


images/airflow/3.0.6/python/mwaa/entrypoint.py
```diff
@@ -261,6 +261,12 @@ async def main() -> None:
     executor_type = os.environ.get("MWAA__CORE__EXECUTOR_TYPE", "CeleryExecutor")
     environ = setup_environment_variables(command, executor_type)
 
+    # Install RDS IAM credential rotation for non-migrate-db commands.
+    # migrate-db uses admin credentials for schema migrations.
+    if command != "migrate-db":
```
chgao28: It seems like there was some refactoring done by not utilizing `airflow_local_settings.py`, and instead importing the patch here. However, if I recall correctly, installing the SQL Alchemy listener here won't actually correctly attach to the processes as they are spun up as sub-processes. The only reliable way it to use `airflow_local_settings.py`, since airflow imports this file early on for all Airflow processes.

The explicit `install_rds_iam_patch()` function is good - but it should be invoked from `airflow_local_settings.py` rather than `entrypoint.py`

Keep in mind that if we continue to use `airflow_local_settings.py`, then the relevent changes for copying customer's `airflow_local_settings,py` will need to be added as well.

images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
```diff
@@ -0,0 +1,111 @@
...
+
+from mwaa.config.rds_iam_credentials import (
+    RDSIAMCredentialProvider,
+    is_using_rds_proxy,
+    use_iam_credentials,
+)
+
+logger = logging.getLogger(__name__)
+
+MWAA_AIRFLOW_COMPONENT = os.environ.get("MWAA_AIRFLOW_COMPONENT")
```
chgao28: nit: check for MWAA_AIRFLOW_COMPONENT within the _is_from_migrate_db(), as it isn't used anywhere else. This module-level constant forces the importlib.reload() gymnastics in the new tests.

images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
```diff
@@ -0,0 +1,111 @@
...
+        return
+
+    if not is_using_rds_proxy():
+        logger.info("RDS IAM patch not installed: not using RDS Proxy.")
+        return
+
+    if _is_from_migrate_db():
+        logger.info("RDS IAM patch not installed: running in migrate-db process.")
+        return
+
```
chgao28, line 128: is_local_runner() check seems to be dropped in 3.x

images/airflow/3.0.6/python/mwaa/database/migrate_with_downgrade.py
```diff
@@ -53,28 +58,86 @@ def _connect_static():
...
-                    print(f"Creating user '{DB_IAM_USERNAME}'")
+                    logger.info(f"Creating user '{DB_IAM_USERNAME}'")
                     conn.execute(text(f"CREATE USER {DB_IAM_USERNAME}"))
-                    print(f"Created db rds iam user")
+                    logger.info("Created db rds iam user")
                 else:
-                    print(f"db rds iam user already exists")
+                    logger.info("db rds iam user already exists")
 
                 # Always ensure permissions are up to date
```
chgao28: User check - if the `current_role == 'adminuser'` check is completely skipped - technically, some of these grant statements can only be executed by `adminuser`, and it would just silently fail because `airflow_user` does not have certain privileges

images/airflow/3.0.6/python/mwaa/database/migrate_with_downgrade.py
```diff
@@ -53,28 +58,86 @@ def _connect_static():
...
                 conn.execute(text(f"GRANT rds_iam TO {DB_IAM_USERNAME}"))
```
chgao28: Also missing `GRANT adminuser TO airflow_user` which allows `airflow_user` be a member of the `adminuser` - crucial for fixing a bug that `airflow_user` sometimes don't have permissions to certain tables.

chgao28 (changes requested):

images/airflow/3.0.6/python/mwaa/entrypoint.py
Mercury2699: You were right, and this was the root cause. Fixed in 0f08132.

`airflow_local_settings.py` is restored for all three versions and copied to `${AIRFLOW_USER_HOME}/config` by the Dockerfile; it invokes `install_rds_iam_patch()`, and the call has been removed from `entrypoint.py`. Confirmed the mechanism is unchanged in Airflow 3: `settings.py` appends `$AIRFLOW_HOME/config` to `sys.path` and `import_local_settings()` runs from `settings.initialize()`, which `airflow/__init__.py` calls on every `import airflow`.

Also confirmed your diagnosis about subprocesses: `execute_command.py` launches the components via `Subprocess` -> `subprocess.Popen(["airflow", ...])`, a fresh interpreter, so the listener registered in the entrypoint never reached any process that talks to the metadata DB.

Customer local settings loading is included: `dags/airflow_local_settings.py` and `plugins/airflow_local_settings.py` are loaded exactly as in 2.x.

images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
Mercury2699: Done in 23ae6c4. `MWAA_AIRFLOW_COMPONENT` is now read inside `_is_from_migrate_db()` and the module-level constant is gone (verified it was not referenced anywhere else). All `importlib.reload()` calls are removed from the patch tests, replaced by a small `_reset_patch_state()` helper for the one-shot install flag.

images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
Mercury2699: Correct, it had been dropped. Restored in 0f08132: `is_local_runner()` is added to `rds_iam_credentials.py` alongside the other helpers and is now one of the install conditions.

Worth flagging: `MWAA_LOCAL_RUNNER` is not forwarded into containers by `docker-compose` in either 2.x or 3.x, so this guard is inert inside containers. The effective local protection is `MWAA__DB__POSTGRES_SSLMODE: "prefer"`, which makes `is_using_rds_proxy()` false. I restored the guard and set the variable in `run.sh` for parity, but the sslmode is what actually does the work locally.

images/airflow/3.0.6/python/mwaa/database/migrate_with_downgrade.py
Mercury2699: Fixed in 0f08132. `DB_ADMIN_USERNAME` is restored and the grants are gated on `SELECT current_user` being `adminuser`, with the `elif current_role == DB_IAM_USERNAME` branch, matching 2.x.

images/airflow/3.0.6/python/mwaa/database/migrate_with_downgrade.py
Mercury2699: Fixed in 0f08132. `GRANT adminuser TO airflow_user` is restored at the end of the adminuser branch, with a comment noting it exists so `airflow_user` inherits ownership of objects created by `adminuser` during migrations.

images/airflow/3.0.6/airflow_local_settings.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+import importlib.util
+import logging
+import os
+
+# Airflow copies every public name of this module into the `airflow.settings`
+# namespace. Declaring an empty `__all__` keeps our internals from clobbering
+# Airflow's own globals; we export no policies or settings overrides.
+__all__: list[str] = []
```
chgao28, line 25: This line seems to be new, and I'm ensure what the side effect of this is. If we can ensure that the patch is working correctly, then it should be fine.

chgao28 (dismissed):

images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py
```diff
@@ -0,0 +1,154 @@
...
+listener registered here only affects the process that registers it, so it has
+to be installed there rather than from the entrypoint, whose Airflow components
+run as separate subprocesses.
+"""
+
+import logging
+import os
+from functools import cache
+
+from sqlalchemy.engine import make_url
```
chgao28, line 25: nit: This is `from sqlalchemy import make_url` in 3.2.1. It was also this before a recent commit: `5a19b4bc1007ae3aeae184864f05758d58b51d62`.

From Gemini: 

> Both from sqlalchemy.engine import make_url and from sqlalchemy import make_url import the exact same function.

> from sqlalchemy import make_url is the preferred public API.

images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py
```diff
@@ -141,6 +141,11 @@ def patch_rds_iam_authentication(dialect, conn_rec, cargs, cparams):
         token = RDSIAMCredentialProvider.get_token()
         url = make_url(RDSIAMCredentialProvider.create_db_connection_url(token))
 
+        # Clear existing connection params to avoid conflicts between
+        # SQLAlchemy versions (SA 2.0 uses 'dbname' in cparams while
+        # translate_connect_args returns 'database' -- both present causes
+        # psycopg2 to raise "can't specify both 'database' and 'dbname'").
+        cparams.clear()
```
chgao28: cparams.clear() wipes all pre-existing connect args, not just credentials — so anything Airflow or MWAA injects via engine connect_args (e.g.,
  connect_timeout, keepalive settings, application_name, options) is silently dropped for metadata DB connections; only what's in the IAM URL (host/port/db/user/password/sslmode) survives.

We should confirm if any arguments are potentially affected. Otherwise, it's safer to `pop()` these values before updating the url to prevent unintended effects

chgao28 (changes requested):

images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py
Mercury2699: 3.0.6 uses SA 1.4, so this is the same as in 2.x, no issue.

images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py
Mercury2699: Will use pop to remove the conflicting keys only.

chgao28 (approved):

kuyperse (approved):

PR #591, opened 2026-08-11
fix(images): Match 2.x logging in the RDS IAM patch

Mercury2699:
## Summary

Restores the 2.x logging behaviour of the RDS IAM patch in two parts:

1. **Resolve the metadata database URL at patch-install time** instead of on the first database connection. This stops an internal INFO record from being prepended to MWAA CLI stdout, which breaks JSON parsing for CLI callers.

2. **Remove the five log records** that report the patch was not installed or was already installed. 2.x logs nothing on either path, and these records go to customer log groups.

## Changes

- `images/airflow/3.0.6/python/mwaa/config/rds_iam_patch.py`
- `images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py`
- `images/airflow/3.3.0/python/mwaa/config/rds_iam_patch.py`
- `tests/images/airflow/3.0.6/python/mwaa/config/test_rds_iam_patch_3_0_6.py`
- `tests/images/airflow/3.2.1/python/mwaa/config/test_rds_iam_patch_3_2_1.py`
- `tests/images/airflow/3.3.0/python/mwaa/config/test_rds_iam_patch_3_3_0.py`

## Testing

- Added `test_install_resolves_metadata_url_eagerly` to all three version test files.
- The test asserts `get_db_connection_string()` is called exactly once during `install_rds_iam_patch()` (at install time, not lazily on first connection) and is not called again by a later `_is_accessing_metadata_db()` check.

Port of internal CR-296267560.

### DIFF
--- a/images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
+++ b/images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
@@ -125,7 +125,7 @@ def install_rds_iam_patch() -> None:
     if is_local_runner():
         return
 
-    # Resolve here rather than on the first connection, as 2.x did: it keeps
+    # Resolve here rather than on the first connection to keep
     # get_db_connection_string()'s INFO record out of MWAA CLI stdout.
     _get_metadata_url()
 
--- a/images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
+++ b/images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
@@ -111,24 +111,23 @@ def install_rds_iam_patch() -> None:
     global _patch_installed
 
     if _patch_installed:
-        logger.debug("RDS IAM patch already installed in this process.")
         return
 
     if not use_iam_credentials():
-        logger.info("RDS IAM patch not installed: USE_IAM_CREDENTIALS is not true.")
         return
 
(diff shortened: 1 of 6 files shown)


images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
```diff
@@ -111,25 +111,24 @@ def install_rds_iam_patch() -> None:
...
 
     if _is_from_migrate_db():
-        logger.info("RDS IAM patch not installed: running in migrate-db process.")
         return
 
     if is_local_runner():
-        logger.info("RDS IAM patch not installed: running in local runner mode.")
         return
 
+    # Resolve here rather than on the first connection, as 2.x did: it keeps
```
kuyperse: nit: don't need to mention that this is the same as in 2.x.

kuyperse (dismissed):

chgao28 (dismissed):

images/airflow/3.2.1/python/mwaa/config/rds_iam_patch.py
Mercury2699: True.

chgao28 (approved):

kuyperse (approved):

PR #624, opened 2026-09-18
feat(auth-manager): Wire IDC_AUTH_ENABLED to MwaaAuthManager in airflow.py

raminv80:
When IDC_AUTH_ENABLED=true, sets AIRFLOW__CORE__AUTH_MANAGER to `aws_mwaa.auth_manager.MwaaAuthManager` so that Airflow loads the correct auth manager for IAM Identity Center-based authentication. Previously, FabAuthManager was always loaded regardless of this setting.

Also adds startup validation that raises EnvironmentError with a clear message if required configuration is missing when IDC_AUTH_ENABLED=true (IDC_CLIENT_SECRET_ARN or IDC_CLIENT_ID/IDC_CLIENT_SECRET, IDC_START_URL, and IDC_IDENTITY_STORE_ID), rather than silently proceeding with the wrong auth manager.


wangyimosquito (dismissed):

images/airflow/3.2.1/python/mwaa/config/airflow.py
```diff
@@ -218,9 +179,48 @@ def _get_opinionated_airflow_db_config() -> Dict[str, str]:
...
+
+        _IDC_REQUIRED_VARS = [
+            "IDC_START_URL",
+            "IDC_IDENTITY_STORE_ID",
+        ]
+        missing = [v for v in _IDC_REQUIRED_VARS if not os.environ.get(v)]
+        if missing:
+            raise EnvironmentError(
+                f"IDC_AUTH_ENABLED=true but the following required IDC configuration "
+                f"variables are missing: {missing}. This indicates a service bug — "
```
kuyperse: Is this error customer facing? If so it needs to have more careful wording. i.e. not something like "there is a service bug"

raminv80: Good point, will update the messages.

kuyperse (approved):

wangyimosquito (approved):

